PR #28, opened 2022-06-09
Add support for Xilinx Pynq-Z2 board.

simone-machetti:


core-v-mini-mcu.core
```diff
@@ -54,11 +54,13 @@ filesets:
 
   rtl-fpga:
     files:
+    - hw/fpga/core_v_mini_mcu_wrapper.sv
```
davideschiavone, line 57: rename the file as 
xilinx_core_v_mini_mcu_wrapper as well as the module name

hw/fpga/core_v_mini_mcu_wrapper.sv
```diff
@@ -0,0 +1,54 @@
+// Copyright 2022 OpenHW Group
+// Solderpad Hardware License, Version 2.1, see LICENSE.md for details.
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+
+module core_v_mini_mcu_wrapper #(
```
davideschiavone, line 5: rename it as written above

hw/fpga/core_v_mini_mcu_wrapper.sv
```diff
@@ -0,0 +1,54 @@
...
+    input  logic fetch_enable_i,
+    output logic exit_value_o,
+    output logic exit_valid_o
+);
+
+  logic             clk_gen;
+  logic      [31:0] exit_value;
+
+
+  clk_wizard_wrapper clk_wizard_wrapper_i (
```
davideschiavone, line 31: pls call this module

xilinx_clk_wizard_wrapper_gen_0 

and the module name as

xilinx_clk_wizard_wrapper_i

hw/fpga/scripts/generate_clk_wizard.tcl
```diff
@@ -0,0 +1,119 @@
+# Proc to create BD clk_wizard
```
davideschiavone, line 1: fix this file to remove the too many checks as well as the name as above

hw/fpga/scripts/generate_clk_wizard.tcl
```diff
@@ -0,0 +1,119 @@
...
+  # Set parent object as current
+  current_bd_instance $parentObj
+
+
+  # Create interface ports
+
+  # Create ports
+  set clk_125MHz [ create_bd_port -dir I -type clk -freq_hz 125000000 clk_125MHz ]
+  set clk_out1_0 [ create_bd_port -dir O -type clk clk_out1_0 ]
+  set_property -dict [ list \
```
davideschiavone, line 75: please, set the 20MHz clock frequency in a variable (e.g. set CLK_FREQ_HZ 20000000) at the beginning of the file, and then use the variable in the script

hw/fpga/scripts/generate_clk_wizard.tcl
```diff
@@ -0,0 +1,119 @@
...
+   CONFIG.FREQ_HZ {20000000} \
+ ] $clk_out1_0
+
+  # Create instance: clk_wiz_0, and set properties
+  set clk_wiz_0 [ create_bd_cell -type ip -vlnv xilinx.com:ip:clk_wiz:6.0 clk_wiz_0 ]
+  set_property -dict [ list \
+   CONFIG.CLKIN1_JITTER_PS {80.0} \
+   CONFIG.CLKOUT1_JITTER {172.798} \
+   CONFIG.CLKOUT1_PHASE_ERROR {96.948} \
+   CONFIG.CLKOUT1_REQUESTED_OUT_FREQ {20} \
```
davideschiavone, line 85: same here, please use the variable as much as you can so that the script become easily updatable when we want to change the CLK Frequency

hw/fpga/scripts/generate_clk_wizard.tcl
```diff
@@ -0,0 +1,119 @@
...
+  close_bd_design $design_name
+}
+# End of cr_bd_clk_wizard()
+cr_bd_clk_wizard ""
+set_property REGISTERED_WITH_MANAGER "1" [get_files clk_wizard.bd ]
+set_property SYNTH_CHECKPOINT_MODE "Hierarchical" [get_files clk_wizard.bd ]
+
+#call make_wrapper to create wrapper files
+if { [get_property IS_LOCKED [ get_files -norecurse clk_wizard.bd] ] == 1  } {
+  import_files -fileset sources_1 [file normalize "${origin_dir}/tapeout/core-v-mini-mcu/build/openhwgroup.org_systems_core-v-mini-mcu_0/pynq-z2-vivado/openhwgroup.org_systems_core-v-mini-mcu_0.gen/sources_1/bd/clk_wizard/hdl/clk_wizard_wrapper.v" ]
```
davideschiavone, line 115: fix path with relative path (e.g. .openhwgroup.org_systems_core-v-mini-mcu_0.gen/sources_1/bd/clk_wizard/hdl/clk_wizard_wrapper.v)

the path I used may be wrong, it's an example

davideschiavone: what is probably better is to delete the wrapper in case exists and re-create it all the time so that you do not need CLEAN commands to update the wrapper

davideschiavone (changes requested):

hw/fpga/scripts/xilinx_generate_clk_wizard.tcl
```diff
@@ -0,0 +1,50 @@
+
+# Define design macros
+set design_name                 xilinx_clk_wizard
+set in_clk_freq_hz              125000000
+set out_clk_freq_hz             20000000
```
davideschiavone, line 5: same, for format reasons better to define it as 20.0 and then calculate it with expressions in tcl

hw/fpga/scripts/xilinx_generate_clk_wizard.tcl
```diff
@@ -0,0 +1,50 @@
...
+set clkin1_jitter_ps            80.0
```
davideschiavone, line 6: do we need this number or we can leave it unset and use the default values? do they change from one Vivado version to the other one?

hw/fpga/scripts/xilinx_generate_clk_wizard.tcl
```diff
@@ -0,0 +1,50 @@
...
+set clkout1_jitter              172.798
```
davideschiavone, line 7: do we need this number or we can leave it unset and use the default values? do they change from one Vivado version to the other one?

hw/fpga/scripts/xilinx_generate_clk_wizard.tcl
```diff
@@ -0,0 +1,50 @@
...
+set clkout1_phase_error         96.948
```
davideschiavone, line 8: do we need this number or we can leave it unset and use the default values? do they change from one Vivado version to the other one?

hw/fpga/scripts/xilinx_generate_clk_wizard.tcl
```diff
@@ -0,0 +1,50 @@
...
+set clkout1_requested_out_freq  20
+set mmcm_clkfbout_mult_f        8.000
```
davideschiavone, line 10: do we need this number or we can leave it unset and use the default values? do they change from one Vivado version to the other one?

hw/fpga/scripts/xilinx_generate_clk_wizard.tcl
```diff
@@ -0,0 +1,50 @@
...
+set mmcm_clkin1_period          8.000
```
davideschiavone, line 11: we do not need this value as a variable, please use it directly in the creation of the clk wizard - we want to use as variable only what we may want to change

hw/fpga/scripts/xilinx_generate_clk_wizard.tcl
```diff
@@ -0,0 +1,50 @@
...
+set mmcm_clkout0_divide_f       50.000
+set prim_in_freq                125.000
+set use_locked                  false
```
davideschiavone, line 14: we do not need this value as a variable, please use it directly in the creation of the clk wizard - we want to use as variable only what we may want to change

hw/fpga/scripts/xilinx_generate_clk_wizard.tcl
```diff
@@ -0,0 +1,50 @@
...
+set use_reset                   false
```
davideschiavone, line 15: we do not need this value as a variable, please use it directly in the creation of the clk wizard - we want to use as variable only what we may want to change

PR #30, opened 2022-06-13
External ports interface, resolve #15

### LINKED ISSUE #15: add external master and slave plugs to bus

benoitdenkinger:
Started external peripherals and master/slave ports interface implementation.


davideschiavone:
> I added the external interfaces. I still have some modifications to do to take into account the mapping of the external device (start/end addresses, ...). How do you want the interfaces to be tested? Do you want dummy external peripherals (like copying data from one memory location to another) and master/slave in the testbench?

Good question, we can do a fake master writing to a fake slave (e.g. we can move the slow memory example out of the core v mcu and instantiate it in the testharness). the peripheral interface would serve as to co figure the fake master slave. For what concerns the memory mapping, just pick a large enough memory space and leave it as "external" in the json file, then further bus dispatching will be solved externally to the repository

benoitdenkinger:
For now, I kept the dispatching inside and I'm concatenating the external and system (internal) map rules.

davideschiavone:
@benoitdenkinger , I think we can do this in a more clean way if we have N external masters as you did, but only one external slave and one external peripheral. This way you do not need to pass any detailed information about the address decoder, etc. You simply gather into a big memory space all the requests that go to the external slave and another portion of memory for the external peripheral. This way you implement externally a second stage decoder that further dispatch the requests coming from the external slave. I think this is probably the best way in terms of extendability - if the synthesizer does a good job, maybe it's even equivalent to have a single decoder as now. For the masters instead having only one port would reduce the input bandwidth, so it's ok to have several of those - which do not required anyway any decoding information

hw/core-v-mini-mcu/core_v_mini_mcu.sv
```diff
@@ -67,8 +68,8 @@ module core_v_mini_mcu
   obi_resp_t        debug_slave_resp;
   obi_req_t         peripheral_slave_req;
   obi_resp_t        peripheral_slave_resp;
-  obi_req_t         slow_ram_slave_req;
-  obi_resp_t        slow_ram_slave_resp;
+  // obi_req_t         slow_ram_slave_req;
```
davideschiavone, line 71: please delete things you do not use in the final version of the Pull Request

hw/core-v-mini-mcu/core_v_mini_mcu.sv
```diff
@@ -150,8 +151,8 @@ module core_v_mini_mcu
       .debug_slave_resp_i     (debug_slave_resp),
       .peripheral_slave_req_o (peripheral_slave_req),
       .peripheral_slave_resp_i(peripheral_slave_resp),
-      .slow_ram_req_o         (slow_ram_slave_req),
-      .slow_ram_resp_i        (slow_ram_slave_resp),
+      // .slow_ram_req_o         (slow_ram_slave_req),
```
davideschiavone, line 154: please delete things you do not use in the final version of the Pull Request

hw/core-v-mini-mcu/core_v_mini_mcu.sv
```diff
@@ -193,28 +196,28 @@ module core_v_mini_mcu
...
-      .gnt_o(slow_ram_slave_resp.gnt),
-      .rdata_o(slow_ram_slave_resp.rdata),
-      .rvalid_o(slow_ram_slave_resp.rvalid)
-  );
-`else
-  assign slow_ram_slave_resp.gnt = '0;
-  assign slow_ram_slave_resp.rdata = '0;
-  assign slow_ram_slave_resp.rvalid = '0;
-`endif
+  // `ifndef SYNTHESIS
```
davideschiavone, line 199: please delete things you do not use in the final version of the Pull Request

hw/core-v-mini-mcu/include/core_v_mini_mcu_pkg.sv
```diff
@@ -111,12 +111,12 @@ package core_v_mini_mcu_pkg;
   localparam logic [31:0] UART_IDX = 32'd1;
 
 
-  localparam addr_map_rule_t [SYSTEM_NPERIPHERALS-1:0] PERIPHERALS_ADDR_RULES = '{
+  localparam addr_map_rule_t [SYSTEM_NPERIPHERALS-1:0] SYSTEM_PERIPHERALS_ADDR_RULES = '{
```
davideschiavone, line 114: after the last comment, we can revert these changes

hw/core-v-mini-mcu/system_bus.sv
```diff
@@ -51,11 +51,11 @@ module system_bus
     output obi_req_t  peripheral_slave_req_o,
     input  obi_resp_t peripheral_slave_resp_i,
 
-    output obi_req_t  slow_ram_req_o,
-    input  obi_resp_t slow_ram_resp_i,
+    // output obi_req_t  slow_ram_req_o,
```
davideschiavone, line 54: please delete things you do not use in the final version of the Pull Request

hw/core-v-mini-mcu/system_bus.sv
```diff
@@ -92,7 +92,7 @@ module system_bus
   assign ram1_req_o = slave_req[core_v_mini_mcu_pkg::RAM1_IDX];
   assign debug_slave_req_o = slave_req[core_v_mini_mcu_pkg::DEBUG_IDX];
   assign peripheral_slave_req_o = slave_req[core_v_mini_mcu_pkg::PERIPHERAL_IDX];
-  assign slow_ram_req_o = slave_req[core_v_mini_mcu_pkg::SLOW_MEMORY_IDX];
+  // assign slow_ram_req_o = slave_req[core_v_mini_mcu_pkg::SLOW_MEMORY_IDX];
```
davideschiavone, line 95: please delete things you do not use in the final version of the Pull Request

tb/testharness.sv
```diff
@@ -85,4 +111,24 @@ module testharness #(
       .exit           ()
   );
 
+  assign slow_ram_slave_req = slave_req[testharness_pkg::SLOW_MEMORY_IDX];
```
davideschiavone, line 114: perfect, that's the idea, and this is also what actually happens externally, and if you need more slaves you add a bus as in the system_bus module

hw/core-v-mini-mcu/core_v_mini_mcu.sv
```diff
@@ -2,10 +2,14 @@
 // Solderpad Hardware License, Version 2.1, see LICENSE.md for details.
 // SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
 
-module core_v_mini_mcu #(
+module core_v_mini_mcu
+  import obi_pkg::*;
+  import addr_map_rule_pkg::*;
```
davideschiavone, line 7: why do we need this package?

hw/core-v-mini-mcu/include/core_v_mini_mcu_pkg.sv
```diff
@@ -111,12 +121,12 @@ package core_v_mini_mcu_pkg;
...
       '{idx: SOC_CTRL_IDX, start_addr: SOC_CTRL_START_ADDRESS, end_addr: SOC_CTRL_END_ADDRESS},
       '{idx: UART_IDX, start_addr: UART_START_ADDRESS, end_addr: UART_END_ADDRESS}
   };
 
-  localparam int unsigned PERIPHERALS_PORT_SEL_WIDTH = SYSTEM_NPERIPHERALS > 1 ? $clog2(
+  localparam int unsigned SYSTEM_PERIPHERALS_PORT_SEL_WIDTH = SYSTEM_NPERIPHERALS > 1 ? $clog2(
```
davideschiavone, line 129: same, why did you add SYSTEM?

hw/core-v-mini-mcu/peripheral_external.sv
```diff
@@ -0,0 +1,102 @@
+// Copyright 2022 OpenHW Group
```
davideschiavone, line 1: if this is an example IP that goes to the testharness, I suggest to call add it to the example IP folder

hw/core-v-mini-mcu/peripheral_external.sv
```diff
@@ -0,0 +1,102 @@
...
+    input  obi_req_t  slave_req_i,
+    output obi_resp_t slave_resp_o,
+
+    output obi_req_t  master_req_o,
+    input  obi_resp_t master_resp_i
+);
+
+  import testharness_pkg::*;
+  import reg_pkg::*;
+  import tlul_pkg::*;
```
davideschiavone, line 20: why you need tlul?

hw/core-v-mini-mcu/peripheral_external.sv
```diff
@@ -0,0 +1,102 @@
...
+// Solderpad Hardware License, Version 2.1, see LICENSE.md for details.
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+
+module peripheral_external
+  import obi_pkg::*;
+(
+    input logic clk_i,
+    input logic rst_ni,
+
+    input  obi_req_t  slave_req_i,
```
davideschiavone, line 11: actually, the external peripheral bus should come already from the peripheral bus, i.e. the input of the external peripheral should not be an obi_req_t bus, but it should be a req_req_t input as done  here: https://github.com/davideschiavone/core-v-mini-mcu/blob/main/hw/core-v-mini-mcu/peripheral_subsystem.sv#L96.

Otherwise this could go together with the external slave

hw/core-v-mini-mcu/peripheral_external.sv
```diff
@@ -0,0 +1,102 @@
...
+      .rst_ni,
+      .in_select_i(peripheral_select),
+      .in_req_i(peripheral_req),
+      .in_rsp_o(peripheral_rsp),
+      .out_req_o(peripheral_slv_req),
+      .out_rsp_i(peripheral_slv_rsp)
+  );
+
+  // Example peripheral with master port to access memory
+  memcopy_periph #(
```
davideschiavone, line 86: only the memcopy_periph should be here, as all the above should be done by the system - see comment above

mcu_cfg.hjson
```diff
@@ -28,9 +28,14 @@
         }
     },
 
-    slow_memory: {
+    ext_peripherals: {
```
davideschiavone, line 31: in my opinion, the peripheral should be under _**peripherals**_

like uart and soc_ctrl

hw/core-v-mini-mcu/include/core_v_mini_mcu_pkg.sv
```diff
@@ -90,9 +95,14 @@ package core_v_mini_mcu_pkg;
           end_addr: PERIPHERAL_END_ADDRESS
       },
       '{
-          idx: SLOW_MEMORY_IDX,
-          start_addr: SLOW_MEMORY_START_ADDRESS,
-          end_addr: SLOW_MEMORY_END_ADDRESS
+          idx: EXT_PERIPHERAL_IDX,
```
davideschiavone, line 98: the external peripheral should go together with uart and soc_ctrl, the slow memory instead will be on the external slave - see also comments below

tb/testharness.sv
```diff
@@ -33,10 +36,32 @@ module testharness #(
...
+  obi_resp_t [testharness_pkg::EXT_XBAR_NMASTER-1:0] master_resp;
+  obi_req_t slave_req;
+  obi_resp_t slave_resp;
+  obi_req_t periph_slave_req;
+  obi_resp_t periph_slave_resp;
+
+  obi_req_t slow_ram_slave_req;
+  obi_resp_t slow_ram_slave_resp;
+
+  // assign master_req[0].req = 0;
```
davideschiavone, line 50: please remove commented code in the final commint

hw/core-v-mini-mcu/core_v_mini_mcu.sv
benoitdenkinger: Yes, I need to clean to code. Thanks for the detailed feedback

hw/core-v-mini-mcu/core_v_mini_mcu.sv
benoitdenkinger: That's an old relic of the original implementation I did. I need to clean these changes that are not needed anymore

hw/core-v-mini-mcu/peripheral_external.sv
benoitdenkinger: Removed

hw/core-v-mini-mcu/core_v_mini_mcu.sv
benoitdenkinger: removed

hw/core-v-mini-mcu/core_v_mini_mcu.sv
benoitdenkinger: removed

hw/core-v-mini-mcu/include/core_v_mini_mcu_pkg.sv
benoitdenkinger: Reverted

hw/core-v-mini-mcu/system_bus.sv
benoitdenkinger: Removed

tb/testharness.sv
benoitdenkinger: ok

hw/core-v-mini-mcu/include/core_v_mini_mcu_pkg.sv
```diff
@@ -111,12 +121,12 @@ package core_v_mini_mcu_pkg;
   localparam logic [31:0] UART_IDX = 32'd1;
 
 
-  localparam addr_map_rule_t [SYSTEM_NPERIPHERALS-1:0] PERIPHERALS_ADDR_RULES = '{
+  localparam addr_map_rule_t [SYSTEM_NPERIPHERALS-1:0] SYSTEM_PERIPHERALS_ADDR_RULES = '{
```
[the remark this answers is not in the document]
benoitdenkinger: Reverted

hw/core-v-mini-mcu/include/core_v_mini_mcu_pkg.sv
benoitdenkinger: Reverted

hw/core-v-mini-mcu/peripheral_external.sv
benoitdenkinger: Inside the hw/ip_examples/memcopy_periph?

hw/core-v-mini-mcu/peripheral_external.sv
benoitdenkinger: But I need to decode the address in case there are multiple external peripherals (it's not the case with the current example), no? Otherwise, I need something like I did before where I give the core_v_min_cpu.sv the mapping rule for all the external peripheral.

tb/testharness.sv
benoitdenkinger: Removed

hw/core-v-mini-mcu/peripheral_external.sv
davideschiavone: thanks indeed - otherwise it is a copy of the slave port

hw/core-v-mini-mcu/peripheral_external.sv
davideschiavone: well, if you do not add any other external peripheral then you do not need to decode it - if you add two peripherals then yes

benoitdenkinger: Or do you want to keep the example simple with only one external peripheral for now?

davideschiavone: yes it works for me - as you want

hw/core-v-mini-mcu/peripheral_external.sv
benoitdenkinger: I removed it and directly instantiated the memcopy peripheral inside the testharness and removed the external peripheral bus decoder as we have only one external peripheral.

hw/core-v-mini-mcu/peripheral_external.sv
benoitdenkinger: Done

hw/core-v-mini-mcu/peripheral_external.sv
benoitdenkinger: I simplified the example by removing the decoding stage (which was useless in this case).

hw/core-v-mini-mcu/include/core_v_mini_mcu_pkg.sv
benoitdenkinger: done

tb/testharness.sv
```diff
@@ -85,4 +117,33 @@ module testharness #(
       .exit           ()
   );
 
+  assign slow_ram_slave_req = slave_req;
+  assign slave_resp = slow_ram_slave_resp;
+
+  slow_memory #(
```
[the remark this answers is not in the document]
benoitdenkinger: done

benoitdenkinger:
I implemented all your comments. I also tested it with the NtoM bus type and it works.

sw/Makefile
```diff
@@ -19,17 +19,17 @@
...
-RISCV                    ?= ~/.riscv
+RISCV		    ?= ~/.riscv
 RISCV_EXE_PREFIX         = $(RISCV)/bin/riscv32-unknown-elf-
 
 # GCC configuration
 CUSTOM_GCC_FLAGS           =
 
-LIB_CRT                    = $(wildcard device/lib/crt/*.S)
-LIB_BASE                   = $(wildcard device/lib/base/*.c)
+LIB_CRT		    = $(wildcard device/lib/crt/*.S)
```
davideschiavone, line 28: why the spacing are changing here?

sw/device/lib/drivers/memcopy_periph.h
```diff
@@ -0,0 +1,60 @@
+// Copyright 2022 OpenHW Group
```
davideschiavone, line 1: as above

sw/device/lib/drivers/memcopy_periph_regs.h
```diff
@@ -0,0 +1,35 @@
+// Generated register defines for memcopy_periph
```
davideschiavone, line 1: as above

tb/testharness.sv
```diff
@@ -127,13 +133,35 @@ module testharness #(
...
+      .obi_resp_t(obi_pkg::obi_resp_t)
+  ) memcopy_periph_i (
+      .clk_i,
+      .rst_ni,
+      .reg_req_i(memcopy_periph_req),
+      .reg_rsp_o(memcopy_periph_rsp),
+      .master_req_o(master_req[testharness_pkg::EXT_MASTER0_IDX]),
       .master_resp_i(master_resp[testharness_pkg::EXT_MASTER0_IDX])
   );
+// `else
```
davideschiavone: pls remove the commented code below

davideschiavone:
@benoitdenkinger - I asked for a couple of minor changes (just moving some files to the mem copy application folder and invert the semantic of the flag).
One last comment: it would be nice if you can do file .md where you explain what you did for the new peripheral, such that we can document the effort you made and other people can do it independently than the 2 of us

sw/Makefile
benoitdenkinger: Reverted

sw/device/lib/drivers/memcopy_periph.c
```diff
@@ -0,0 +1,28 @@
+// Copyright 2022 OpenHW Group
```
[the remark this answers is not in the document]
benoitdenkinger: done

sw/device/lib/drivers/memcopy_periph.h
benoitdenkinger: done

sw/device/lib/drivers/memcopy_periph_regs.h
benoitdenkinger: done

tb/testharness.sv
benoitdenkinger: done

sw/Makefile
benoitdenkinger: done

benoitdenkinger:
I also had to modify the regtool.py file to make it work:
@@ -218,7 +218,7 @@ def main():
         if found_lunder:
             src_lic = found_lunder
         if found_spdx:
-            src_lic += '\n' + found_spdx
+            src_lic = str(src_lic) + '\n' + str(found_spdx)
 
         with outfile:
             if format == 'html':

Is it working for you, or should I do a patch?

davideschiavone:
> I also had to modify the regtool.py:
> 
>     * src_lic += '\n' + found_spdx
> 
> 
>     * src_lic = str(src_lic) + '\n' + str(found_spdx)
> 
> 
> Is it working for you, or should I do a patch?

where did you modify it?

davideschiavone:
btw, the CI checks Verible and it failed

benoitdenkinger:
In hw/vendor/pulp_platform_register_interface/vendor/lowrisc_opentitan/util/regtool.py

benoitdenkinger:
I added the md file. Let me know if that's what you had in mind.

benoitdenkinger:
> I also had to modify the regtool.py:
> 
>     * src_lic += '\n' + found_spdx
> 
> 
>     * src_lic = str(src_lic) + '\n' + str(found_spdx)
> 
> 
> Is it working for you, or should I do a patch?

Never mind, the problem was in my hjson file, you have to use the lowrisc copyright header otherwise the script fails.

I pushed a new version

Issue #20, opened 2022-06-06
add PLIC


davideschiavone:
Can be vendorzed from: https://github.com/lowRISC/opentitan/tree/master/hw/top_earlgrey/ip_autogen/rv_plic

see: https://docs.opentitan.org/hw/top_earlgrey/ip_autogen/rv_plic/doc/

benoitdenkinger:
I'll check this one

benoitdenkinger:
What should I connect to the PLIC for now? The uart interrupts? Should I use the external memcopy as an example to generate an interrupt? So if someone wants to add its own peripheral, there is one example.

davideschiavone:
Hi @benoitdenkinger - adding an external-internal bus like you did with masters is definitively a good idea.

In addition, we want to connect to PLIC the peripherals as in Opentitan.

CLINT RISC-V defines that some of the interrupts go directly into the core (e.g. software, timer), and the others go to a PLIC, which gets N events, and forwards 1 interrupt (called external) to the CPU - see the specs and also here: https://github.com/lowRISC/opentitan/blob/master/hw/top_earlgrey/rtl/autogen/top_earlgrey.sv#L2134

benoitdenkinger:
Ok, but now all the peripherals are not integrated (only uart). Do we let the responsibility to everyone integrating a peripheral with interrupts to connect it to the plic or should I wait for more peripheral to be integrated (SPI, DMA, ...)?

davideschiavone:
1st option, so connect only UART as for now

benoitdenkinger:
At the software level, does the interrupt handler check the interrupt source or the handler only notify there is an interrupt pending (i.e., should I read the plic registers inside the interrupt handler or outside to minimize the interrupt time)?

davideschiavone:
As you can see from here: https://github.com/lowRISC/opentitan/blob/master/hw/top_earlgrey/rtl/autogen/top_earlgrey.sv#L2134

the PLIC only produces one interrupt, which goes to the external interrupt of the CPU, thus, in ISR does not have any knowledge about which of the PLIC sources generated the interrupt. So yes, you need to read the registers of the PLIC

benoitdenkinger:
Yes I understood that, my question was not clear enough. The question is: do I read the register inside the ISR? Or do a set a flag to 1 in the ISR (so the ISR is very fast) and then in my application I first check the flag and then read the registers if needed?

davideschiavone:
done in #39

PR #39, opened 2022-06-30
PLIC integration (issue #20)

benoitdenkinger:
I integrated the plic. I kept the default 32 interrupt vector size even if most of them are not used, they will be in near future. I tested it with modelsim/questsim. With verilator I have weird errors when I try to compile. For example:
ERROR: %Error: ../../../hw/vendor/lowrisc_opentitan/hw/lint/tools/verilator/common.vlt:11: syntax error, unexpected '-'
lint_off -rule PINCONNECTEMPTY
It complains about the dash (which is correct). So I couldn't test it with verilator for now.


benoitdenkinger:
I forgot to check one thing, so please wait before accepting the pull request

benoitdenkinger:
I pushed another commit. I modified the vector table and add weak function handlers (similar to opentitan). I think it simplifies the integration of other interrupts. I tested all the examples and they work properly with questasim. I still have problems with verilator, so if someone could test the external peripheral example with it that would be great.

PR #41, opened 2022-07-04
Timer

simone-machetti:
Added hw and sw support for rv_timer module.


davideschiavone:
you need to solve the conflicts for this PR to be reviewd

PR #45, opened 2022-07-05
rv_timer

simone-machetti:
Added hw and sw support for rv_timer module.


sw/device/lib/drivers/rv_timer/rv_timer.c
```diff
@@ -0,0 +1,372 @@
+// Copyright EPFL contributors.
```
davideschiavone, line 1: this file seems to be taken from lowRisc, thus you should not change the licence

davideschiavone (changes requested):

PR #48, opened 2022-07-11
Added hw and sw support for gpio module.

simone-machetti:


davideschiavone (changes requested):
1. fix rv_plic.c
2. fix Xilinx wrapper and pin assignment
3. add GPIO toggling via timers & switch tu use both the timer and the GPIO interrupts

davideschiavone (approved):

Issue #35, opened 2022-06-23
High-level Configuration


davideschiavone:
- [ ] Make RAM configurable, including LD script
- [ ] Make ID of master and slaves auto inferred by the tool
- [ ] Delete? generated-by-template files and include their generation from fusesoc with mcu gen
- [ ] Make Peripherals included or not by selection
- [ ] Interactive script to select CPU, MEM SIZE, and BUS type

davideschiavone:
RAM and LD script done in #119, but leaving it unflagged as the cut size is fixed to 32KB

davideschiavone:
Generate PAD Ring in #123

davideschiavone:
WIP https://github.com/esl-epfl/x-heep/pull/605

PR #53, opened 2022-07-22
CPU and BUS high-level configuration - resolve partially #35

benoitdenkinger:
I created a Makefile for most tools. Now CPU and BUS can be used to specify a different value than the ones in the mcu-cfg.hjson file. I checked everything was working with verilator and questasim.


davideschiavone:
hi @benoitdenkinger thanks a lot, I will try it out asap, mea while, can you please update the README and DEBUG file ? and also, how can one add the flags to fusesoc like use external device or jtag dpi?

benoitdenkinger:
Hi @davideschiavone, yes sure. I didn't consider all the different cases, only the main ones. I can move everything to the Makefile and put in the README only the corresponding make command for all the different tools and options.

davideschiavone:
it probably makes sense to use a make variable "FUSESOC_FLAGS" where make passes to fusesoc

benoitdenkinger:
I updated the README and the Makefile. I kept the previous instructions (how to compile for the different tools with fusesoc) and added a section related to the Makefile. I also removed the $ sign from the commands to have an easier copy-paste of the commands.

davideschiavone:
you may want to change also the ExternalDevice.md

.gitignore
```diff
@@ -4,3 +4,6 @@ build/
 *.so
 *.hex
 *.elf
+
+# ignore core-v-mini-mcu pkg that is automatically generated by Makefile
+hw/core-v-mini-mcu/include/core_v_mini_mcu_pkg.sv
```
davideschiavone, line 9: shouldn't we do the same with SW?

davideschiavone (changes requested):

.gitignore
benoitdenkinger: Yes, you are right.

benoitdenkinger:
Simulation with verilator works but not with modelsim. In modelsim, the instruction address loop over the first few bootrom instructions (h0 to h10, it does not jump to <_jump_to_flash>).

davideschiavone:
> Simulation with verilator works but not with modelsim. In modelsim, the instruction address loop over the first few bootrom instructions (h0 to h10, it does not jump to <_jump_to_flash>).

I still have to write a readme for executing from Flash,  there is an open issued. does modelsim work without the boot from Flash flag?

benoitdenkinger:
My previous comment was not correct/precise. I didn't try to run from flash so I guess it's normal it does not jump to _jump_to_flash. But it doesn't work for modelsim even without trying to boot from flash

davideschiavone:
> My previous comment was not correct/precise. I didn't try to run from flash so I guess it's normal it does not jump to _jump_to_flash. But it doesn't work for modelsim even without trying to boot from flash

ok then tomorrow we have a look

davideschiavone:
this function is probably not called in tb_top.sv

tb_set_exit_loop

whereas it is called in the tb_top.cpp

tb/tb_top.sv
```diff
@@ -55,28 +57,46 @@ module tb_top #(
...
+        boot_sel = 0;
+      end
+    end else begin
+      $display("[TESTBENCH]: No Boot Option specified, using jtag");
+      boot_sel = 0;
+    end
+
+    wait (rst_n == 1'b1);
+
+    if (JTAG_DPI == 0 || boot_sel == 1) begin
```
davideschiavone, line 91: why are we pre-loading when booting from flash?

davideschiavone: I would say "if(JTAG_DPI == 0 && boot_sel = 0)

davideschiavone: and else if (boot_sel == 0)

tb/tb_top.sv
```diff
@@ -55,28 +57,46 @@ module tb_top #(
...
+      testharness_i.tb_loadHEX(firmware);
+      #CLK_PHASE_HI
+      testharness_i.tb_set_exit_loop();
+      #CLK_PHASE_LO
+        if ($test$plusargs("verbose")) $display("[TESTBENCH] %t: memory loaded", $time);
+    end else begin
```
davideschiavone, line 95: end else if (boot_sel == 0) begin

davideschiavone: I fix it

PR #57, opened 2022-07-25
DMA implementation - resolve #13 (add new DMA design)

### LINKED ISSUE #13: add DMA
- [ ]  add new DMA design
- [ ] connect DMA to SPI
- [ ] connect DMA to GPIO to wake it up on transfers
- [ ] connect TIMER to DMA

benoitdenkinger:
DMA implementation using two masters (one for reading and one for writing) and a FIFO of 4 between them. Tested with NtoM bus type to assess maximum speed is achieved when reading from ram1 and writing to ram0.


hw/ip/dma/rtl/dma.sv
```diff
@@ -0,0 +1,279 @@
...
+
+  // DMA pulse start when dma_start register is written
+  always_ff @(posedge clk_i or negedge rst_ni) begin : proc_dma_start
+    if (~rst_ni) begin
+      dma_start <= 1'b0;
+    end else begin
+      if (dma_start == 1'b1) begin
+        dma_start <= 1'b0;
+      end else begin
+        dma_start <= |reg2hw.dma_start.q;
```
davideschiavone, line 108: this is SW mapped, so it means that the CPU in SW needs to write to 1 and then to 0 right? does it make sense?

hw/ip/dma/rtl/dma.sv
```diff
@@ -0,0 +1,279 @@
...
+  assign data_out_gnt = dma_master1_ch0_resp_i.gnt;
+  assign data_out_rvalid = dma_master1_ch0_resp_i.rvalid;
+  assign data_out_rdata = dma_master1_ch0_resp_i.rdata;
+
+  assign dma_intr_o = dma_done;
+
+  assign hw2reg.done.de = dma_done | dma_start;
+  assign hw2reg.done.d = dma_done == 1'b1 ? 1'b1 : 1'b0;
+
+  assign hw2reg.dma_start.de = dma_start;
```
davideschiavone, line 97: ah sorry didnt see it, indeed you did it ;)

hw/ip/dma/data/dma.hjson
```diff
@@ -0,0 +1,46 @@
...
+      fields: [
+        { bits: "31:0", name: "PTR_OUT", desc: "Output data pointer" }
+      ]
+    },
+    { name:     "DMA_START",
+      desc:     "Number of bytes to copy - Once a value is written, the copy starts",
+      swaccess: "rw",
+      hwaccess: "hrw",
+      fields: [
+        { bits: "31:0", name: "DMA_START", desc: "DMA counter and start" }
```
davideschiavone, line 33: although more efficient like you did, does it make sense to split start and counter in 2 different registers? I am thinking like you set the counter once, and the you start and stop later, not necessarily at the same time, not saying it's not ok, just reasoning

hw/vendor/lowrisc_opentitan/hw/ip/rv_plic/rtl/rv_plic_reg_top.sv
```diff
@@ -45,10 +46,39 @@ module rv_plic_reg_top (
   tlul_pkg::tl_h2d_t tl_reg_h2d;
   tlul_pkg::tl_d2h_t tl_reg_d2h;
 
-  assign intg_err_o = '0;
+  // incoming payload check
```
davideschiavone, line 49: are you sure it compiles with the intg_err logic back?

hw/ip/dma/data/dma.hjson
benoitdenkinger: Yes, I had exactly the same thinking but didn't find a strong advantage for one or the other. I decided to go for this one as you have to write to start the transfer anyway (and this way you can decide to change the number of bytes copied every time if needed). But maybe one solution will become better once the interface with the SPI is implemented.

hw/ip/dma/rtl/dma.sv
```diff
@@ -144,7 +144,7 @@ module dma #(
       if (dma_start == 1'b1) begin
         dma_cnt <= reg2hw.dma_start.q;
       end else if (data_in_gnt == 1'b1) begin
-        dma_cnt <= dma_cnt - 32'h1;
+        dma_cnt <= dma_cnt - 32'h4;
```
davideschiavone, line 147: from a SW view point, if you can specify "bytes" it means you can write "3", but this won't work

benoitdenkinger:
The new version is ready for review

benoitdenkinger:
I rebased and solved the conflicts.

PR #63, opened 2022-07-28
Power domains

simone-machetti:


sw/device/lib/drivers/power_manager/power_manager.c
```diff
@@ -0,0 +1,137 @@
...
+#include "mmio.h"
+
+#include "power_manager_regs.h"  // Generated.
+
+void power_gate_core(void)
+{
+    asm volatile (
+
+        // write POWER_GATE_CORE[0] = 1
+        "lui a0,0x20030\n"
```
davideschiavone, line 19: you should use the DEFINEs and not hard numbers

sw/device/lib/drivers/power_manager/power_manager.c
```diff
@@ -0,0 +1,137 @@
...
+        "addi a1, a0, 0\n"
+        "li  a3,1\n"
+        "sw  a3,0(a1)\n"
+
+        // write WAKEUP_STATE[0] = 1
+        "lui a0,0x20030\n"
```
davideschiavone, line 25: you should use the DEFINEs and not hard numbers and so on

PR #86, opened 2022-08-15
Ext interrupt (fix #67)

### LINKED ISSUE #67: fix the value of EXT_NINTERRUPT
The EXT_NINTERRUPT parameter theoretically allows anyone to add any number of external interrupts.

However, they must be between the number of current interrupts that the PLIC support (41) and 64.

So I suggest removing this parameter and fixing it to 64-41 = 23.

If the external user needs less, he/she can simply tight to 0 the ones unused.

If she/he wants more, at least it is known it is not possible without modifying the x-heep hw and sw platform.

benoitdenkinger:
I removed the EXT_INTERRUPT from the testharness, but kept it in the top core_v_mini_mcu.sv file to keep it simple to change.

### DIFF
--- a/hw/core-v-mini-mcu/core_v_mini_mcu.sv
+++ b/hw/core-v-mini-mcu/core_v_mini_mcu.sv
@@ -10,7 +10,7 @@ module core_v_mini_mcu
     parameter FPU = 0,
     parameter PULP_ZFINX = 0,
     parameter EXT_XBAR_NMASTER = 0,
-    parameter EXT_NINTERRUPT = 0
+    parameter EXT_NINTERRUPT = 6
 ) (
     input logic clk_i,
     input logic rst_ni,
--- a/hw/core-v-mini-mcu/core_v_mini_mcu.sv
+++ b/hw/core-v-mini-mcu/core_v_mini_mcu.sv
@@ -9,8 +9,7 @@ module core_v_mini_mcu
     parameter PULP_XPULP = 0,
     parameter FPU = 0,
     parameter PULP_ZFINX = 0,
-    parameter EXT_XBAR_NMASTER = 0,
-    parameter EXT_NINTERRUPT = 6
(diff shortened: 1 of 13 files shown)


hw/core-v-mini-mcu/core_v_mini_mcu.sv
```diff
@@ -10,7 +10,7 @@ module core_v_mini_mcu
     parameter FPU = 0,
     parameter PULP_ZFINX = 0,
     parameter EXT_XBAR_NMASTER = 0,
-    parameter EXT_NINTERRUPT = 0
+    parameter EXT_NINTERRUPT = 6
```
davideschiavone, line 13: can we move it to localparam instead ? 

something like

localparam PLIC_NINT = 64;
localparam PLIC_USED_NINT = 57:
localparam NEXT_INT = PLIC_NINT - PLIC_USED_NINT + 1;

they should go the the core v mcu pkg such that external chips (e.g. heepocrates) can read this number

benoitdenkinger: I moved the paramater to the mcu top package and added them to the mcu hjson file to generate them with the template

mcu_cfg.hjson
```diff
@@ -77,7 +77,10 @@
 
 
     interrupts: {
-        // always zero
+        plic_intr_src:           64,
```
davideschiavone, line 80: are you sure it makes sense to have it exposed to the user? I do not think the user can change it anyway

benoitdenkinger: My idea was to put it here in case it needs to be changed it can be done in the json file and not in the RTL files.

davideschiavone: but if you change it to 32 or 128 this won't change the RTL, so it is actually dangerous to expose it to the users

PR #94, opened 2022-08-19
SPI OT SW and test (fix #12)

### LINKED ISSUE #12: add SPI and/or QSPI
SPI can be vendorized from https://github.com/lowRISC/opentitan/tree/master/hw/ip/spi_host, see https://docs.opentitan.org/hw/ip/spi_host/doc/

- [x] add HW
- [ ] add SW
- [ ] test it

benoitdenkinger:
SPI OT SW implementation with interrupt tested.


hw/ip/boot_rom/boot_rom.S
```diff
@@ -2,47 +2,156 @@
...
+       // Wait spi is ready for next command
+       lw     a4, SPI_HOST_STATUS_REG_OFFSET(a1)
+       bgez   a4, _wait_spi_ready_cmd_pwr
+
+       // Fill TX FIFO with TX data (read command 0x03 + 3B address 0x000)
+       li     a4, 0x03
+       sw     a4, SPI_HOST_DATA_REG_OFFSET(a1)
+
+_wait_spi_ready_tx_init:
+       nop    # otherwise ready bit check is too fast
```
davideschiavone, line 90: I don't really like this point, I will investigate but it does not look safe

hw/ip/boot_rom/boot_rom.S
```diff
@@ -2,47 +2,156 @@
...
+       lw     a4, SPI_HOST_STATUS_REG_OFFSET(a1)
+       bgez   a4, _wait_spi_ready_tx_init
+       // Read command: 0x00002203
+       lui    a4, 0x2
+       addi   a4, a4, 515 # spi cmd: txonly + stdspeed + csaat + 4B
+       sw     a4, SPI_HOST_COMMAND_REG_OFFSET(a1)
+
+_wait_spi_ready_read_prog:
+       nop    # otherwise ready bit check is too fast
```
davideschiavone, line 99: as above

hw/ip/boot_rom/boot_rom.S
```diff
@@ -2,47 +2,156 @@
...
+       bgez   a5, _wait_spi_ready_read_32B_chunk
+       addi   s7, s1, 256 # add 32*8 (256 bytes) to dst ptr
+
+_wait_spi_rxwm_8_words:
+       lw     a5, SPI_HOST_STATUS_REG_OFFSET(a1)
+       srli   a5, a5, 0x14
+       andi   a5, a5, 1 # SPI_HOST_STATUS_RXWM_BIT
+       beqz   a5, _wait_spi_rxwm_8_words
+       // Read 32B (8 words) from RX FIFO
+       lw     a7, SPI_HOST_DATA_REG_OFFSET(a1)
```
davideschiavone, line 132: although less efficient, a loop would decrease the ROM size, especially as these LW and SW are not compressed

hw/ip/soc_ctrl/data/soc_ctrl.hjson
```diff
@@ -52,12 +52,21 @@
...
         { bits: "0", name: "USE_SPIMEMIO", desc: "Spi Select Reg" }
       ]
     }
+    { name:     "ENABLE_SPI_SEL",
+      desc:     "Enable Spi module selection from software",
+      resval:   "0x0"
+      swaccess: "rw",
+      hwaccess: "hro",
+      fields: [
+        { bits: "0", name: "ENABLE_SPI_SEL", desc: "Enable Spi Select Reg" }
```
davideschiavone, line 67: why do we want to enable the previous bit? is that what you want to do?

hw/ip/soc_ctrl/rtl/soc_ctrl.sv
```diff
@@ -41,7 +43,7 @@ module soc_ctrl #(
   assign hw2reg.boot_select.de = 1'b1;
   assign hw2reg.boot_select.d = boot_select_i;
 
-  assign hw2reg.use_spimemio.de = 1'b1;
+  assign hw2reg.use_spimemio.de = ~enable_spi_sel;
```
davideschiavone, line 46: ah ok got it

hw/ip/soc_ctrl/data/soc_ctrl.hjson
benoitdenkinger: I implemented this because at the beginning I used the spiflash crt0 that is using the spimemio to copy the data from flash to ram, so I needed a way to switch between the two SPI from software. Now I only rely on the OT SPI host, so this could be removed, but I believe it is still interesting to be able to switch from one to the other after the booting is done.

PR #100, opened 2022-08-25
fix #99

davideschiavone:

PR #102, opened 2022-08-26
Added macro for uart baudrate by default based on the target

JoseCalero:


davideschiavone:
already done here: https://github.com/esl-epfl/x-heep/pull/100

PR #110, opened 2022-08-31
add power manager and cpu pd

davideschiavone:
- [x] check verilator
- [x] check modelsim with upf
- [x] check design compiler
- [x] check bitstream generation


davideschiavone:
replace #63

PR #140, opened 2022-10-10
fix flash linker and model

davideschiavone:


davideschiavone:
pls @ruben-roalvarez check you as well

PR #143, opened 2022-10-10
Fix heep tb core and verilator

davideschiavone:


davideschiavone:
by mistake I branched the linker script commit... just merge this after #140

PR #153, opened 2022-10-17
Add write and read to flash in the spi-dma example, plus fix spi-flash interrupt

ruben-roalvarez:


sw/applications/spi_flash_write/spi_flash_write.c
```diff
@@ -0,0 +1,274 @@
...
+#include "rv_plic.h"
+#include "rv_plic_regs.h"
+#include "soc_ctrl.h"
+#include "spi_host.h"
+#include "dma.h"
+#include "fast_intr_ctrl.h"
+#include "fast_intr_ctrl_regs.h"
+
+// Un-comment this line to use the SPI FLASH instead of the default SPI
+#define USE_SPI_FLASH
```
benoitdenkinger, line 22: Can you comment the USE_SPI_FLASH to be consistent with the comment above

sw/applications/spi_flash_write/spi_flash_write.c
```diff
@@ -0,0 +1,274 @@
...
+{
+    // Disable SPI interrupts
+    spi_enable_evt_intr(&spi_host, false);
+    spi_enable_rxwm_intr(&spi_host, false);
+    // Clear fast interrupt
+    fast_intr_ctrl_t fast_intr_ctrl;
+    fast_intr_ctrl.base_addr = mmio_region_from_addr((uintptr_t)FAST_INTR_CTRL_START_ADDRESS);
+    clear_fast_interrupt(&fast_intr_ctrl, kSpiFlash);
+    spi_intr_flag = 1;
+}
```
benoitdenkinger, line 54: Can you but them in an ifndef with USE_SPI_FLASH for a better understanding (except if both are actually needed in both cases)

sw/applications/spi_flash_write/spi_flash_write.c
```diff
@@ -0,0 +1,274 @@
...
+    // Enable SPI host device
+    spi_set_enable(&spi_host, true);
+
+    // SPI and SPI_HOST are the same IP so same register map
+    uint32_t *fifo_ptr = spi_host.base_addr.base + SPI_HOST_DATA_REG_OFFSET;
+
+    // -- SPI CONFIGURATION -- 
+    // Configure chip 0 (flash memory)
+    // Max 50 MHz core SPI clock --> Max 25 MHz SCK
+    // Single data IO; keep timing as safe as possible.
```
benoitdenkinger, line 97: The TB runs @100MHz by default. Is there any limit on the FLASH side for the max frequency? If this is the case, you can check the actual frequency in the SoC controller and set the clock divider accordingly.

sw/applications/spi_flash_write/spi_flash_write.c
```diff
@@ -0,0 +1,274 @@
...
+        .speed      = kSpiSpeedStandard,
+        .direction  = kSpiDirTxOnly
+    });
+
+    const uint32_t cmd_dummy_read = spi_create_command((spi_command_t){
+        .len        = 2,
+        .csaat      = false,
+        .speed      = kSpiSpeedStandard,
+        .direction  = kSpiDirRxOnly
+    });
```
benoitdenkinger, line 124: Please add a small comment about the need for these dummy commands. Consider renaming cmd_dummy to cmd_dummy_write as you have cmd_dummy_read.

sw/applications/spi_flash_write/spi_flash_write.c
```diff
@@ -0,0 +1,274 @@
...
+    spi_write_word(&spi_host, powerup_byte_cmd);
+    const uint32_t cmd_powerup = spi_create_command((spi_command_t){
+        .len        = 0,
+        .csaat      = false,
+        .speed      = kSpiSpeedStandard,
+        .direction  = kSpiDirTxOnly
+    });
+    spi_set_command(&spi_host, cmd_powerup);
+    spi_wait_for_ready(&spi_host);
+    printf("Wait...\n");
```
benoitdenkinger, line 154: Consider removing printf that do not provide so much information and slow down the simulation

sw/applications/spi_flash_write/spi_flash_write.c
```diff
@@ -0,0 +1,274 @@
...
+    spi_write_word(&spi_host, write_enable_cmd);
+    const uint32_t cmd_write_en = spi_create_command((spi_command_t){
+        .len        = 0,
+        .csaat      = false,
+        .speed      = kSpiSpeedStandard,
+        .direction  = kSpiDirTxOnly
+    });
+    spi_set_command(&spi_host, cmd_write_en);
+    spi_wait_for_ready(&spi_host);
+    printf("Wait...\n");
```
benoitdenkinger, line 171: Same than previous

sw/applications/spi_flash_write/spi_flash_write.c
```diff
@@ -0,0 +1,274 @@
...
+    #endif
+    CSR_SET_BITS(CSR_REG_MIE, mask);
+    spi_intr_flag = 0;
+
+    // Select SPI host as SPI output
+    soc_ctrl_select_spi_host(&soc_ctrl);
+    // Enable SPI host device
+    spi_set_enable(&spi_host, true);
+
+    // SPI and SPI_HOST are the same IP so same register map
```
benoitdenkinger, line 91: // SPI and SPI_FLASH are the...

sw/applications/spi_flash_write/spi_flash_write.c
```diff
@@ -0,0 +1,274 @@
...
+            .csaat      = false,
+            .speed      = kSpiSpeedStandard,
+            .direction  = kSpiDirBidir
+        });
+        spi_set_command(&spi_host, spi_bi_cmd);
+        spi_wait_for_ready(&spi_host);
+        spi_wait_for_rx_watermark(&spi_host);
+        spi_read_word(&spi_host, &flash_resp[0]);
+        if ((flash_resp[0] & 0x01) == 0) flash_busy = false;
+    }
```
benoitdenkinger, line 247: Can't you use the SPI interrupt for that?

sw/applications/spi_flash_write/spi_flash_write.c
```diff
@@ -0,0 +1,274 @@
...
+
+    // Fix bug with Bidirectional additional byte
```
benoitdenkinger, line 249: the cmd_dummy_read is not RX only (so not bidirectional)?

sw/applications/spi_flash_write/spi_flash_write.c
```diff
@@ -0,0 +1,274 @@
...
+    spi_set_command(&spi_host, cmd_dummy_read);
+    spi_wait_for_ready(&spi_host);
+    spi_wait_for_rx_watermark(&spi_host);
+    spi_read_word(&spi_host, &flash_resp[0]);
+
+    printf("Power down flash...\n");
```
benoitdenkinger, line 255: Please check that you actually write to the flash by reading back the values

sw/applications/spi_host_dma_example/spi_host_dma_example.c
```diff
@@ -68,28 +69,24 @@ int main(int argc, char *argv[])
...
-        dma_set_spi_mode(&dma, (uint32_t) 3); // The DMA will wait for the SPI rx FIFO valid signal
+        dma_set_spi_mode(&dma, (uint32_t) 3); // The DMA will wait for the SPI FLASH RX FIFO valid signal
     #endif
 
-    // ---------------------
-
+    // -- SPI CONFIGURATION -- 
     // Configure chip 0 (flash memory)
     // Max 50 MHz core SPI clock --> Max 25 MHz SCK
     // Single data IO; keep timing as safe as possible.
```
benoitdenkinger, line 92: Remove this and check if there is any limitation on the max SPI clock for the Flash.

sw/applications/spi_host_dma_example/spi_host_dma_example.c
```diff
@@ -105,22 +102,42 @@ int main(int argc, char *argv[])
     spi_set_configopts(&spi_host, 0, chip_cfg);
     spi_set_csid(&spi_host, 0);
 
+    const uint32_t cmd_dummy = spi_create_command((spi_command_t){
```
benoitdenkinger, line 105: Add comment about the necessity of this command

sw/applications/spi_host_dma_example/spi_host_dma_example.c
```diff
@@ -105,22 +102,42 @@ int main(int argc, char *argv[])
...
     spi_wait_for_ready(&spi_host);
+    spi_set_command(&spi_host, cmd_dummy);
+    spi_wait_for_ready(&spi_host);
 
     // The address bytes sent through the SPI to the Flash are in reverse order
     uint32_t flash_data_addr = flash_data;
-    uint32_t read_byte_cmd = (flash_data_addr >> 16);
+    uint32_t read_byte_cmd = (flash_data_addr >> 16); 
     read_byte_cmd = (read_byte_cmd | (flash_data_addr & 0xff00));
     read_byte_cmd = (read_byte_cmd | ((flash_data_addr & 0xff) << 16));
```
benoitdenkinger, line 142: Define a macro (so it can be re-used easily)

sw/applications/spi_host_dma_example/spi_host_dma_example.c
```diff
@@ -142,25 +159,39 @@ int main(int argc, char *argv[])
     spi_wait_for_ready(&spi_host);
 
     const uint32_t cmd_read_rx = spi_create_command((spi_command_t){
-        .len        = SPI_BYTES - 1,
+        .len        = 31, // SPI_BYTES - 1,
```
benoitdenkinger, line 162: Remove unnecessary comments

sw/applications/spi_host_example/spi_host_example.c
```diff
@@ -15,6 +15,9 @@
 #include "fast_intr_ctrl.h"
 #include "fast_intr_ctrl_regs.h"
 
+// Un-comment this line to use the SPI FLASH instead of the default SPI
+#define USE_SPI_FLASH
```
benoitdenkinger, line 19: Comment it be default

sw/applications/spi_host_example/spi_host_example.c
```diff
@@ -36,10 +39,28 @@ void handler_irq_fast_spi(void)
...
+    spi_enable_evt_intr(&spi_host, false);
+    spi_enable_rxwm_intr(&spi_host, false);
+
+    // Clear fast interrupt
+    fast_intr_ctrl_t fast_intr_ctrl;
+    fast_intr_ctrl.base_addr = mmio_region_from_addr((uintptr_t)FAST_INTR_CTRL_START_ADDRESS);
+    clear_fast_interrupt(&fast_intr_ctrl, kSpiFlash);
+
+    spi_intr_flag = 1;
+}
```
benoitdenkinger, line 54: Put it in an ifndef with USE_SPI_FLASH and the handler_irq_fast_spi above function

sw/applications/spi_host_example/spi_host_example.c
```diff
@@ -48,7 +69,11 @@ int main(int argc, char *argv[])
     // Enable global interrupt for machine-level interrupts
     CSR_SET_BITS(CSR_REG_MSTATUS, 0x8);
     // Set mie.MEIE bit to one to enable machine-level fast spi interrupt
-    const uint32_t mask = 1 << 20;
+    #ifndef USE_SPI_FLASH
+        const uint32_t mask = 1 << 20;
+    #else
+        const uint32_t mask = 1 << 21;
+    #endif
```
benoitdenkinger, line 76: Please also remove the comment about the max 50 MHz frequency below.

benoitdenkinger:
@ruben-roalvarez The SPI write example code has been tested only for the FPGA implementation with a real flash chip? There is no support for simulation, correct?

ruben-roalvarez:
Yes, the write to flash example has only been tested with the FPGA version

sw/applications/spi_flash_write/spi_flash_write.c
ruben-roalvarez: I don't think so, once all the data has been sent to the FLASH the interrupt triggers and you have to check the status register of the FLASH until it tells you that the page has already been programmed correctly (it takes some time). This is not on the SPI side but on the FLASH side. Pls, correct me if I am missing something

sw/applications/spi_flash_write/spi_flash_write.c
ruben-roalvarez: I meant the previous transaction, which is bidirectional and I think it has a bug as well because I was getting an additional 00 byte when I was performing a reading with the write and read complete application (Davide told me to separate them). I will modify the comment to be clearer

sw/applications/spi_flash_write/spi_flash_write.c
ruben-roalvarez: Davide told me not to perform write and read in the same application because it becomes too long, I am checking by reading the entire FLASH with the iceprog yosys program, I will double-check with him

ruben-roalvarez: @davideschiavone do you agree with this?

ruben-roalvarez: @benoitdenkinger you're right, I will add the readback to check the values

sw/applications/spi_flash_write/spi_flash_write.c
benoitdenkinger: Ok, I see. I actually read the code too fast, now it makes sense

sw/applications/spi_flash_write/spi_flash_write.c
```diff
@@ -0,0 +1,373 @@
...
+    spi_set_enable(&spi_host, true);
+
+    // SPI and SPI_FLASH are the same IP so same register map
+    uint32_t *fifo_ptr = spi_host.base_addr.base + SPI_HOST_DATA_REG_OFFSET;
+
+    // -- SPI CONFIGURATION -- 
+    // Configure SPI clock
+    // SPI clk toggles at half of the freq as the core clk when clk_div = 1
+    // SPI_CLK = CORE_CLK/(2 + 2 * CLK_DIV) < CLK_MAX => CLK_DIV > (CORE_CLK/CLK_MAX - 2)/2
+    const uint16_t clk_div = (REFERENCE_CLOCK_Hz/(1000*1000*FLASH_CLK_MAX) - 2)/2 + 1;
```
benoitdenkinger, line 115: Nice change. Just take the clock from the soc_ctrl peripheral as the REFERENCE_CLOCK_Hz might not be the current frequency (e.g., in an actual chip with an FLL, REF_CLOCK is usually 32768Hz and use to create a faster clock)

PR #173, opened 2022-11-12
Bus fifo

benoitdenkinger:
Added optional peripheral bus fifo to break long timing path.


davideschiavone:
@benoitdenkinger , you gotta run make mcu-gen before pushing

davideschiavone:
replaced by #200

PR #176, opened 2022-11-23
Fix apps and linker.

simone-machetti:


davideschiavone:
@JoseCalero can you pls have a look at it?

simone-machetti:
@JoseCalero Can we merge this branch?

davideschiavone:
one question, shouldn't we have the ALIGN constraints equal to 4096 for the 3 linker scripts?

JoseCalero:
@simone-machetti if you checked that the linker change works for all the other apps (as we were discussing), then we can merge. However, as @davideschiavone is mentioning, that change should be done for all the different linkers. Thus, before doing the merge, @simone-machetti, can you use for instance the on-chip linker with the same app and check this? (To avoid failures by coincidence, just play a little bit with the power manager section location within the linker and also double check the alignment by `readelf -S xxx.elf`)

simone-machetti:
Thanks @JoseCalero.

simone-machetti:
All tests passed! Ready to be merged!

simone-machetti:
@JoseCalero, just double check it, so then Davide merges it. Thank you very much!

PR #184, opened 2022-12-09
Add padrick to generate padframe

Xinhu-L:
This is the draft pr which introduce padrick into X-heep to generate the padframe.


hw/system/padframe/src/padframe.sv
```diff
@@ -0,0 +1,206 @@
...
+  output resp_t                              config_rsp_o
+  );
+
+
+  req_t xheep_config_req;
+  resp_t xheep_config_resp;
+  padframe_xheep #(
+    .req_t(req_t),
+    .resp_t(resp_t)
+  ) i_xheep (
```
davideschiavone, line 87: can u pls call it padframe_xheep_i?

davideschiavone (changes requested):
@Xinhu-L , would it make sense to add in the README or somewhere how to download padrick and generate such padframe ?

Xinhu-L:
Hi Davide, I found that I cannot connect to my computer in the office through ssh today, and I will be back in TU Delft tomorrow, so most likely that I might not be able to regenerate the padframe on my side. However, I have updated the README.md with some instructions. You can modify the name in the padrick configuration YAML file and regenerate it quickly through the instructions (Maybe you also need to change some file names in the driver folder, but it won't be trouble).
Please refer to Michael or Manuel if there is still some trouble, and I will also try to help as much as I can when I am back to TU Delft

JuanSapriza:
Closing because this is no longer needed.

PR #188, opened 2022-12-15
Auto create virtualenv for FuseSoC + other python scripts

meggiman:


davideschiavone:
@christophmuellerorg will love this contribution

PR #189, opened 2022-12-20
add Linux-hosted fpga emulation platform (femu)

Sawii00:
PR to integrate the embedded emulation platform to x-heep.

PR #191, opened 2023-01-05
add FreeRTOS support and CMake compatible sw compilation

JoseCalero:
1. Makefile doc automatisation (self-documenting Makefile, type 'make' and see the output)
2. CMake system integration (riscv integration with its toolchain etc.)
3. FreeRTOS porting (see new linker tpl - need to increase 'flash'/ram0 to fit FreeRTOS .text/code into it, new vectors.S to set new handlers for portASM into FreeRTOS)


davideschiavone:
will be reviewed after #189 is merged

JoseCalero:
@davideschiavone, I think we can merge this

davideschiavone (approved):

PR #192, opened 2023-01-09
add PDM2PCM peripheral

grinningmosfet:


hw/ip/pdm2pcm/rtl/fir.sv
```diff
@@ -0,0 +1,100 @@
...
+// Solderpad Hardware License, Version 2.1, see LICENSE.md for details.
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+
+// Author: Pierre Guillod <pierre.guillod@epfl.ch>, EPFL, STI-SEL
+// Date: 14.12.2022
+// Description: FIR filter with symmetric coefficients
+
+module fir #(
+    // Width of the datapath
+    parameter  WIDTH,
```
davideschiavone, line 11: please add everywhere default parameter values

hw/ip/pdm2pcm/rtl/pdm2pcm.sv
```diff
@@ -0,0 +1,161 @@
...
+      .DEPTH(FIFO_DEPTH)
+  ) pdm2pcm_fifo_i (
+      .clk_i,
+      .rst_ni,
+      .flush_i(reg2hw.control.clear.q),
+      .testmode_i(1'b0),
+      .full_o(full),
+      .empty_o(empty),
+      .usage_o(fifo_usage),
+      .data_i({{14{1'b0}}, pcm_o}),
```
davideschiavone, line 130: why pushing 14b equal to 0? can't you add those 14b at the pop side so that you do not waste flip flops?

davideschiavone (changes requested):

hw/ip/pdm2pcm/rtl/pdm2pcm.sv
grinningmosfet: I need to check the FIFO source code/documentation for a width parameter. I thought that the width was fixed at 32 bits, but it may actually be the default value of a parameter.

hw/ip/pdm2pcm/rtl/fir.sv
grinningmosfet: I'll do this.

hw/ip/pdm2pcm/rtl/pdm2pcm.sv
grinningmosfet:
```
module fifo_v3 #(
    parameter bit          FALL_THROUGH = 1'b0, // fifo is in fall-through mode
    parameter int unsigned DATA_WIDTH   = 32,   // default data width if the fifo is of type logic
    parameter int unsigned DEPTH        = 8,    // depth can be arbitrary from 0 to 2**32
)
```
Yes, it's a default parameter. I'll fix it at the width of the datapath (18 bits) and do the necessary adaptations.

davideschiavone: well, if you push (for example) 4 bits only, and then when you pop you extend the popped value to 18b, wouldn't this mean having (18-4)*FIFO_DEPTH fewer flip flops?

Theoretically, the synthesizer will recognize that those are constant 0s are removing them - but it is a better design practice to design saving area wherever possible IMO

hw/ip/pdm2pcm/rtl/pdm2pcm.sv
```diff
@@ -0,0 +1,161 @@
...
+  logic              [                3:0]     par_decim_idx_combs;
+  logic              [                4:0]     par_decim_idx_hfbd2;
+  logic              [                5:0]     par_decim_idx_fir;
+  logic              [               17:0]     coeffs_hb1          [ 0:3];
+  logic              [               17:0]     coeffs_hb2          [ 0:6];
+  logic              [               17:0]     coeffs_fir          [0:13];
+
+  logic              [FIFO_ADDR_WIDTH-1:0]     fifo_usage;
+
+  logic                                        pcm_data_valid_o;
```
davideschiavone, line 35: this is not an output, please remove `_o` from here and wherever else :) same in case you have internal signals named `_i` as they were inputs

davideschiavone (changes requested):

PR #196, opened 2023-02-03
automatic structs generation

StefanoAlbini96:
Python file to automatically generate header file for the peripherals containing structs to access registers' fields


util/headers_periph_gen.py
```diff
@@ -0,0 +1,74 @@
...
+
+# Generates the header file with the structs for every peripheral specified by a json file
+
+template_path = "./sw/device/lib/drivers/template.tpl"     # template used for the output
+json_files = []                                             # list of all the json files
+output_files = []                                           # list of all the output file names
+peripheral_names = []                                       # names of the peripherals for which the structs are generated
+
+# dma
+json_files.append("./hw/ip/dma/data/dma.hjson")
```
davideschiavone, line 11: why instead of hardwiring the hjson files in the script you do not create a HJSON/TXT or whatever file that lists all the HJSON files you need to take as input and pass it to the script? then the script iterates over the HJSON files listed in the array?

davideschiavone (changes requested):

util/headers_periph_gen.py
JoseCalero: Davide is right, let's use the mcu_cfg.hjson and add there the fields for every peripheral @StefanoAlbini96

JoseCalero (changes requested):
use the mcu_cfg.hjson and add there the fields for every peripheral

util/structs_periph_gen.py
```diff
@@ -0,0 +1,78 @@
...
+    ao_peripherals = data["ao_peripherals"]
+    peripherals = data["peripherals"]
+
+    # loops through the peripherals and add the useful ones
+    scan_peripherals(ao_peripherals)
+    scan_peripherals(peripherals)
+
+    # Manually add rv_plic and spi_host due to name differences between mcu_cfg.hjson and peripherals folders #
+    #rv_plic
+    JSON_FILES.append("./hw/vendor/lowrisc_opentitan/hw/ip/rv_plic/data/rv_plic.hjson")
```
davideschiavone, line 63: why is plic not taken from the hjson?

util/structs_periph_gen.py
```diff
@@ -0,0 +1,78 @@
...
+    OUTPUT_FILES.append("./sw/device/lib/drivers/rv_plic/rv_plic_structs.h")
+    PERIPHERAL_NAMES.append("rv_plic")
+
+    # spi_host
+    JSON_FILES.append("./hw/vendor/lowrisc_opentitan_spi_host/data/spi_host.hjson")
+    OUTPUT_FILES.append("./sw/device/lib/drivers/spi_host/spi_host_structs.h")
```
davideschiavone, line 69: why is the spi not taken from the hjson?

davideschiavone (changes requested):

util/structs_periph_gen.py
StefanoAlbini96: Because in the hjson they have different names (plic and spi) with respect to the folder in which the structs need to be generated (rv_plic and spi_host). With @JoseCalero we decided to hardcode them instead of adding an additional "name" field in the hjson.

davideschiavone: but can't you just put the path in the path field where the json file is? I don't get what you need, the SW one can be hardwired in the script or pass the `./sw/device/lib/drivers/` as an argument, the rest is generated by the name of the peripheral no?

StefanoAlbini96: Yes I can add their path to the hjson.

StefanoAlbini96: Yes I can add the path to the hjson. For the output folder I have already hardcoded the base path `./sw/device/lib/drivers/` in the script, but the rest of the path cannot be generated with the name since the names are different, So I would need to hardcode them anyway

PR #202, opened 2023-02-17
Automate interrupt integration

fimtrey:
Hi @davideschiavone,

I was skimming through the files to get an interrupts and thought some automation might be helpful.

It's only a start. Do you have an idea how to automate the `peripheral_subsystem.sv`: (as this looks error-prone)
```systemverilog
  //...        
  assign intr_vector[43] = i2c_intr_tx_empty;
  assign intr_vector[44] = i2c_intr_tx_nonempty;
  assign intr_vector[45] = i2c_intr_tx_overflow;
  assign intr_vector[46] = i2c_intr_acq_overflow;
  assign intr_vector[47] = i2c_intr_ack_stop;
  assign intr_vector[48] = i2c_intr_host_timeout;
  assign intr_vector[49] = spi2_intr_event;
  // ...             ^ should be coming from cfg 
```


mcu_cfg.hjson
```diff
@@ -136,57 +136,59 @@
...
-        intr_trans_complete:     42,
-        intr_tx_empty:           43,
-        intr_tx_nonempty:        44,
-        intr_tx_overflow:        45,
-        intr_acq_overflow:       46,
-        intr_ack_stop:           47,
-        intr_host_timeout:       48,
-        spi2_intr_event:         49,
-        ext_intr:                50:63,
+        number: 64,
```
davideschiavone, line 139: can u pls add a comment that this number should not be changed? I hope people do not change it to 32 or 128

davideschiavone (changes requested):

mcu_cfg.hjson
fimtrey: I could move it into the `mcu_gen.py` and double check all that there are no duplicates. What do you think?

davideschiavone: it can stay there that's fine :) it is easier to see thanks

PR #206, opened 2023-02-27
DMA add circular mode

fimtrey:
Changes:
1. Adding Circular mode
2. Renaming of SPI Mode

**In circular mode** 
the dma automatically restarts the transaction, which enables continuous data streaming, by processing the first and the second have of the buffer separately. The interrupt is raised twice per transaction, a first time at halfway and a second at the end of the buffer. The register bit `halfway` indicates in which have the dma is working.

Better names for the `rx/tx_wait_...` signals and modes? 

Should I add a second dedicated interrupt for halfway ?

Is there any specified behaviour for a write to `dma_start` during an ongoing transaction?
It seems to me that some values of the new transaction would be read but the internal fifo not cleared.

Proposition:
- a 0 to `dma_start` => cancel ongoing transaction (maybe reset circular mode)
- a length > 0 to `dma_start` => start new transaction as soon as done.


sw/device/lib/drivers/dma/dma.h
```diff
@@ -67,10 +93,31 @@ void dma_set_read_ptr_inc(const dma_t *dma, uint32_t read_ptr_inc);
...
+ * 1 = spi_rx to mem
+ * 2 = mem to spi_tx
+ * 3 = spi_flash_rx to mem
+ * 4 = mem to spi_flash_tx
+ * 
  * @param dma Pointer to dma_t represting the target DMA.
- * @param spi_mode 0: mem to mem - 1: spi_rx to mem (Default: 0) - 2: mem to spi_tx.
+ * @param spi_mode (Default: 0)
  */
 void dma_set_spi_mode(const dma_t *dma, uint32_t spi_mode);
```
fimtrey, line 122: @davideschiavone I reverted to keep 
```c
void dma_set_spi_mode(const dma_t *dma, uint32_t spi_mode);
```
in the hal as you suggested.

fimtrey:
I added a start pending register, which triggers a new transaction as soon as done is set. Therefore a early write to `dma_start` shouldn't break the system anymore, but writes to `datatype` or `tx/rx_wait_mode` will still influence the current transaction.

fimtrey:
Move to https://github.com/JuanSapriza/x-heep/pull/1

PR #213, opened 2023-03-03
Removed bit-fields from peripherlas' structs generation

StefanoAlbini96:


JoseCalero (changes requested):
Can we have the pointer also auto generated?
e.g.: #define dma_peri  ((dma *) DMA_START_ADDRESS)
(for that we need to `#include "core_v_mini_mcu.h"` also)

JoseCalero (approved):

PR #203, opened 2023-02-22
add i2s peripheral

fimtrey:

PR #219, opened 2023-03-09
add DMA HAL

JuanSapriza:
# :warning: This documentation is deprecated :warning: 
 ## Please refer to [this one](https://github.com/esl-epfl/x-heep/pull/219#issuecomment-1573854939) below.
 
 ---
---

The DMA is the module that can perform memory transfers with minimal CPU interaction, freeing it to perform other tasks/go to sleep. 
Our DMA is capable to make memory transactions from memory to memory, from memory to a peripheral, and from a peripheral to memory. Peripheral to peripheral transactions are not supported. 
Right now, two peripheral interfaces exist: a general purpose SPI and an SPI connected to a flash memory. 
In previous work (hardware included), the name for this peripherals was hardcoded, yet in the implementation of this HAL they are treated in a more abstract way to allow  incorporating new interfaces with the least possible changed needed in the HAL.

### Transaction
A transaction is an operation to be performed by the DMA. It implies copying bytes from a source pointer into a destination pointer. The transaction configuration can be loaded into the DMA registers once it has been cross-checked and it only starts when the size of the transaction is written in its corresponding register. The transaction is finished once the DMA has sent all its bytes (which not necessarily means they have been received by the final destination).  

### Source and destination
Sources and destinations are the two pointers that will exchange data. Bytes will be copied from the source and into the destination address. 

### Data type
The DMA allows for transactions every 1, 2 or 4 Bytes (Byte, Half Word and Word respectively). The size in bytes of the chosen data type is named 'data unit'. For example, 16 bytes can be 16 data units if the data type is Byte, but 8 data units if the data type is Half Word. 

### Increment
In the case that source and/or destination data are not to be consecutively read/written, a certain increment can be defined. 
For instance, If you have an array of 4-bytes words, but only want to copy the first 2 bytes of each word, you could define the transaction with a Data Type of Half Word and an increment of 2 data units in the source, and 1 data unit in the destination. This way, after each read operation the DMA will increment the read pointer in 4 bytes (2 data units), but the write pointer by only 2 bytes.    

### Alignment
When doing transactions with Bytes, the DMA can read/write from any pointer. However, if the Data Type is larger, words should be aligned so the DMA can make a read/write an affect only the chosen bytes. If a Word or Half Word's pointer is not a multiple of 4 or 2 (respectively), the pointer is "misaligned". In some cases the DMA HAL can overcome this problem reducing the data type (which will reflect on an efficiency loss).  

### Environment
An environment is a region of memory that can optionally be defined by the user to let the HAL know that it can read/write from that region. It is useful to make sure the DMA will not affect reserved memory regions.   
Right now, read and write permissions are not supported by environments, meaning that if it is defined, the DMA will be able to read AND write in it.

### Semaphore
If the source or destination pointer is a peripheral, there are lines connecting the peripheral and the DMA that can be used to control the data flow. This lines are named semaphores as they allow/stop the DMA from sending or reading data. Environments and semaphores cannot coexist. 

### Target
A target is either a region of memory or a peripheral to which the DMA will be able to read or write. When targets are pointing to memory, they can be assigned an environment to make sure that they will comply with memory restrictions. 
Targets include a pointer (a point in the memory, or the Rx/Tx buffer in case of peripherals), a size to be copied (if its going to be used as a source), a data type and an increment. Optionally, they can also include a environment to which they belong.   

### Configuration flags 
During the creation or configuration of environments, targets or transactions, there could be inconsistencies or threatening situations (like writing outside the boundaries of a defined region). To provide the user with information about this potentially harmful situations, configuration flags are set while creating each of these entities. They can be unmasked and checked. 
In some cases, when the threat is too risky, a crucial error might be raised and the operation of the configuration is halted. 
If senseless configurations are input to functions, assertions may halt the whole program. This is reserved for extreme situations that mean the program was not properly coded (e.g. a semaphore value that is not among the available ones is passed). 

## Usage
The basic operation of the DMA HAL follows these steps.
* Creating a target
* Creating another target
* Creating a transaction passing one of the targets as source and the other as a destination
* Load the transaction into the DMA 
* Launch the transaction
* Wait for the transaction to be finished. 

Optionally, one or two environments can be created for the targets. The following possibilities exist: 
* Both targets point to memory, 
  * but they dont use an environment
  * one of them uses an environment
  * they both use environments and
    * they are the same
    *  they are different
* Only one of the targets points to memory, 
  * it uses an environment
  *  it doesnt use an environment 

Once the transaction is created, it can be loaded many times. Once its loaded, it can be launched many times. Targets can be used by more than one transaction, both as source or destination. 

## Special cares

There are a number of problems that can be found while configuring a transaction, and the DMA HAL should be able to detect them and warn the user via configuration flags. Please add as comments any other special case found during testing that should be considered.

### Misalignments
If a misalignment is detected, the HAL should warn the user. If it can fix it by using a smaller data type, it should also allow the user go for this option. 

### Discontinuity 
When the increment is larger than the data unit, this may bring trouble. In the case of misalignments, it could be impossible for the DMA to perform the requested operation. A crucial error should be raised and no operation should be allowed. 

### Overflows
Going beyond the boundaries of an environment cannot only be detected during the creation of a target (noticing that the requested pointer + size will exceed the environments' space), but can be a consequence of the combination of targets. For example, both targets can be well inside their environments, but that the choice of copy size and increment make it impossible to copy into the destination without going out bounds. 

### Overridings
The source and destination could be such that while writing the DMA ends up overriding what its reading. This should not happen. However, this is not yet covered as its a complex situation that could be interesting to some applications (like shifting information inside a buffer). Further work is needed in this matter.  

### Inconsistencies
Some configurations are not compatible. For example, it would make no sense to have an increment if the target is a peripheral. Some of these inconsistencies are treated. Some are yet to be discussed. 

## Useful images
There are these two images that sum up the hardware abstraction being done (where environments, targets, etc are portrayed), and an example of the sw flow to make a transaction.  

DMA HAL-Hardware
The DMA and its surroundings systems as perceived by the DMA HAL. Note that targets are merely a pointer to memory or a buffer of a communication interface. 

DMA HAL-Workflow
Example of a typical DMA use case. 

### Allow circularity
In following versions of the DMA, circularity will be possible thanks to changes proposed by @fimtrey. We decided to implement the changes necessary in the DMA HAL later. 

### Multi-transactioning
From the work done by @fimtrey it seems useful to allow the DMA to multiplex between "slow" transactions (such as SPI transfers). This change was deemed non-trivial and was proposed to be targeted later.   

### Use of SPI
The SPI examples that use the DMA do not work if the selected SPI is not FLASH SPI. This did not work even before the DMA HAL changes. 
In the code, the SPI address used is `SPI_START_ADDRESS`, but it is never defined. I tried using `SPI2_START_ADDRESS` and `SPI_HOST_START_ADDRESS` but none of them work. 

* `SPI_START_ADDRESS`: Compilation error
* `SPI2_START_ADDRESS`: The interrupt never arrives
* `SPI_HOST_START_ADDRESS`: Copies 0xffff... everywhere.

I noticed there was one line that was only set if the SPI mode was FLASH:
`soc_ctrl_select_spi_host(&soc_ctrl);` 
However, adding this line did not change the results. 

As this behavior precedes the changes from this HAL, I propose leaving it as future work. 

### Passing interrupt management to the FIC
Right now the interrupt handler from `handlers.c` is overridden by a strong implementation in `dma.c`. It performs some basic/necessary actions and then calls a weak implementation that can be redefined at application level. 
It was proposed to make the fast interrupt controller (FIC) be in charge of the operations performed in it so to stop including interrupt-related modules inside the HAL. 
As this would depend on the merging of the FIC, I propose leaving it as future work.


davideschiavone (changes requested):
util pycache should not be pushed, maybe better to add in ignore, for what concern the HW, you should modify the core_v_mini_mcu.sv.tpl and not the sv file directly, as this is generated - the rest I leave it to @JoseCalero

JuanSapriza:
> util pycache should not be pushed, maybe better to add in ignore, for what concern the HW, you should modify the core_v_mini_mcu.sv.tpl and not the sv file directly, as this is generated - the rest I leave it to JoseCalero

Hmmm... I didn't *knowingly* modified those files. I will double check and try to leave those files out of the PR, because I see my .vscode folder got pushed as well and I have it untracked.

JuanSapriza:
This should be better. 
I untracked some undesired files (didnt add them to the gitignore because they are specific to my IDE). 
I reverted the changes done in hw and other files not DMA-related.

JoseCalero:
I will review it as soon as it is not a draft :)

JuanSapriza:
I leave here the comments we just discussed with @JoseCalero and @StefanoAlbini96

## Documentation Format 
* Adjust comments and function arguments so that everything fits inside the A4 limits
* Fix consistencies in use of white spaces
* Add a second * on comments that are to be included in the documentation
* Divide sections of code with comment blocks and not line comments

## Documentation Content
* Remove unnecessary @ToDo's and fix wrong comments (j_dma, __undef being default)
* Change image: targets are pointers and not regions, this is confusing.

## Code Format
* Divide large _if's_  into local (if-wise) variables.

## Code Content
* For writing full registers, directly write on the *_struct
* To improve function arguments size, usability and avoid passing so many arguments, transform them into a struct. 
* Use a return value and define a default return value. 
* Set the whole _struct to 0 to prevent undesired operations from being launched. 
* Add a define for the masking of the address (magic 3).

## Other things
* Compare code size with and without _static_ before _inline_.

davideschiavone:
please @JuanSapriza test all the APPs :)

JuanSapriza:
> please @JuanSapriza test all the APPs :)

@davideschiavone That is also ongoing.

JuanSapriza:
In the latest push I: 
* Made most of the required changes. I did not add `ret` values, because he structures' flags were already used where a `ret` would be useful.
* I updated the initial description of this PR:
  * Updated the images
  * Added a `Future Work` section. 
* Made the comparison between `static` and/or `inline`: 
  * `inline` functions need to be declared in a header file [[*]](http://www.greenend.org.uk/rjk/tech/inline.html), so other modules would have access to them. I discarded this as a possibility. 
  * Between `static` and `static inline` declarations there was no size change in the generated binary files. This makes me think that the compiler is not "inline-ing" any of these functions. 
  * I would therefore leave them `static inline` as it is the weakest approach (we do not close the door to inline-ing if later we make some modifications and the compiler decides its the best option). 
* Updated all applications using the DMA. Please note that the non-Flash SPI is not working (even before these changes). You can refer to the section `Future work` of the first comment of this PR.

JuanSapriza:
We will first include @fimtrey's changes on circularity and the trigger/slot management. 

This will be done through a PR on my fork. I will merge it, test it and update this PR.

sw/device/lib/drivers/dma/dma.c
```diff
@@ -1,42 +1,827 @@
...
+    uint8_t afterEnv = p_tgt->ptr > p_tgt->env->end;
+    /* If a size was defined. */
+    uint8_t isSize = p_tgt->size_du;
+    /* If the target selected size goes beyond the boundaries of the 
+    environment. Only computed if there is a size defined.*/
+    uint8_t isOutb = isOutbound( p_tgt->ptr, p_tgt->env->end, 
+                                p_tgt->type, p_tgt->size_du, 
+                                p_tgt->inc_du );
+    
+    if( isEnv && ( beforeEnv || afterEnv || (isSize && isOutb) ) )
```
JoseCalero, line 659: break this is two conditions at least

JoseCalero:
Let us know when this is ready for merge consideration :)

JuanSapriza:
These are the things I have yet to do: 

|Task                 | Description                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        | Priority    | Estimated time   |
|---------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------|------------------|
| Slots               |  Check why slots numbered 1,2,4,8 are not working but 1,2,3,4 are (checked with Tim and I can see that in hardware they effectively are correct)                                                                                                                                                                                                                                                                                                                                                                                                                                                   |  High       |  2 h             |
|    Test circular    |                                                                                                                                                                                                                                              Finish testing circular mode with SPI with less than 64 bytes.                                                                                                                                                                                                                                                                                        |   High      | 2 h              |
|      Window         |                                                                                                                                                                                                                                                                                                           Test the window interrupt (require PR of PLIC for this)                                                                                                                                                                                                                                  |    High     |  2 h             |
|     Documentation   | Update the documentation                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           |      Medium | 2 h              |
|Weak implementations |                                                                                                                                                                                                                                                                                                                                                                  Update weak implementations to `optimize("O0")`                                                                                                                                                                                   |   Low       |      1 h         |
|       FIC           |                                                                                                                                                                                                                                                                                                                                                                                                                 Merge PR of FIC and remove the implementations from the DMA                                                                                                                        | Low         | 1 h              |
|      White tests        |                                                                                                                                                                                                                                                                                                                                                                                                                                                                             Repeat all the test I had previously done, updating them to include tests concerning circular mode and window interrupts|       High  |              16 h |

We can discuss if something is missing or if some priorities need to be changed. 
If I dedicated solely to this, it could take a week. 
i will update this comment with my progress. I will change the status of the PR to draft as its clearly not yet ready for merge.

JuanSapriza:
I leave here below an updated version of the documentation (to be exported to the X-HEEP documentation upon merging), as well as a diagram.  Both are works in progress. 

----

# DMA
The **Direct Memory Access (DMA)** peripheral allows data transfers with reduced CPU interaction. 
It can perform *trasactions* of data between peripherals and memory, or memory-to-memory (as a `memcpy` would). 
The CPU is required to configure the transaction, but once launched it is free to go to sleep, process the incoming data or do anything else. 
 
## The DMA HAL
The DMA **Hardware Abstraction Layer (HAL)** facilitates the configuration of transactions from the users application. Furthermore, it adds an additional layer of safety checks to reduce the risk of faulty memory accesses, data override or infinite loops.  


### Previous Definitions
The implementation of this software layer introduced some concepts that need to be understood in order to make proper use of the DMA's functionalities .

#### Transaction
A transaction is an operation to be performed by the DMA. It implies copying bytes from a source pointer into a destination pointer. The transaction configuration can be loaded into the DMA registers once it has been cross-checked and it only starts when the size of the transaction is written in its corresponding register. The transaction is finished once the DMA has sent all its bytes (which not necessarily means they have been received by the final destination).  
Transactions cannot be stopped once they were launched. 
While a transaction is running, new transactions can be validated, but not launched or loaded into the DMA. 
Transactions can be re-launched automatically in `circular mode`. 
Once the transaction has finished, a status bit is changed (that can be monitored through polling) and a fast interrupt is triggered. 

#### Source and destination
Sources and destinations are the two pointers that will exchange data. Bytes will be copied from the source and into the destination address. 

#### Data type
The DMA allows transactions in chunks of 1, 2 or 4 Bytes (`Byte`, `Half-Word` and `Word` respectively). The size in bytes of the chosen data type is called _data unit_ (usually abreviated as `du`). 
For example, 16 bytes can be 16 data units if the data type is `Byte`, but 8 data units if the data type is `Half Word`. 

#### Increment
In the case that source and/or destination data are not to be consecutively read/written, a certain increment can be defined. 
For instance, if you have an array of 4-bytes-words, but only want to copy the first 2 bytes of each word, you could define the transaction with a data type of half word, an increment of 2 data units in the source, and 1 data unit in the destination. This way, after each read operation the DMA will increment the read pointer in 4 bytes (2 data units), but the write pointer by only 2 bytes.    

#### Alignment
When doing transactions with bytes, the DMA can read/write from any pointer. However, if the data type is larger, words should be aligned so the DMA can perform a read/write operation and affect only the chosen bytes. If a word or half-word's pointer is not a multiple of 4 or 2 (respectively), the pointer is _misaligned_. In some cases the DMA HAL can overcome this problem reducing the data type (which will reflect on an efficiency loss).  

#### Environment
An environment is a region of memory that can optionally be defined by the user to let the HAL know that it is allowed to read/write on that region. It is useful to make sure the DMA will not affect reserved memory regions.   
Right now, read and write permissions are not supported by environments, meaning that if it is defined, the DMA will be able to read AND write on it.

#### Triggers and Slots
If the source or destination pointer is a peripheral, there are lines connecting the peripheral and the DMA that can be used to control the data flow (they behave as _triggers_). These lines are connected to _slots_ on the DMA and they allow/stop the DMA from reading/writing data.

#### Target
A target is either a region of memory or a peripheral to which the DMA will be able to read/write. When targets are pointing to memory, they can be assigned an environment to make sure that they will comply with memory restrictions. 
Targets include a pointer (a point in the memory, or the Rx/Tx buffer in case of peripherals), a size to be copied (if its going to be used as a source), a data type and an increment.

#### Configuration flags 
During the creation or configuration of environments, targets or transactions, there could be inconsistencies or threatening situations (like writing outside the boundaries of a defined region). To provide the user with information about this potentially harmful situations, configuration flags are set while creating each of these entities. They can be unmasked and checked. 
In some cases, when the threat is too risky, a _crucial error_ might be raised and the operation of the configuration is halted. 
If senseless configurations are input to functions, assertions may halt the whole program. This is reserved for extreme situations that mean the program was not properly coded (e.g. a slot value is provided and is not among the available ones). 

#### Transaction modes
To take full advantage of the speed and transaprency of the DMA, a _circular_ mode was implemented. When selected, the DMA will relaunch the exactly same transaction upon finishing. This cycle only stops if by the end of a transaction the _transaction mode_ was changed to _single_ (default). 
The CPU receives a fast interrupt on every transaction finished. 

#### Windows
In order to process information as it arrives, the application can define a _window size_ (smaller than the _transaction size_. Every time the DMA has finished sending that given amount of information will trigger an interrupt through the PLIC. 
> :warning: If the window size is a multiple of the transaction size, upon finishing the transaction there will be first an interrupt for the whole transaction (through the FIC), and then an interrupt for the window (through the PLIC, which is slower). 


#### Checks and Validations
The DMA HAL's interface functions perform two types of checks: 
* **Sanity checks**: Make sure that each individual value passed as an argument is reasonable and belongs to the proper domain. This errors will raise an _assertion_ and, depending on how assertions are managed in the application, may result in the program crashing. 
* **Integrity checks**: Arguments are cross-checked to make sure that they abide by the rules of the DMA. If configurations are incompatible, contradictory, or a risk for the programs integrity, warnings are raised through the _configuration flags_. In some special cases, a _critical error_ can be raised, which blocks the deployment of the configuration into the DMA registers. 

> :warning: Integrity checks can be disabled to speed up configuration time. Do this only if you have previously checked the configuration and are sure it will not cause any trouble.  

Checks and validations are performed during the transactions creation, loading and launching. 

A transaction is validated if it went through the creation-checks without raising critical errors.

#### End events
The DMA consideres a certain amount of bytes to have been transferred once it has sent them. It does not wait for a confirmation from the recipient. When a transaction/window is finished the DMA performs a series of event. These may include:
* Changing its status register.
* Raising a _transaction done_ interrupt.
* Raising a _window done_ interrupt.

The DMA HAL can follow up on these changes or let the application be in charge. For this purpose, three different types of _end events_ are defined: 
* **Polling**: The HAL will disable interrupts from the DMA. The application will need to frequently query the status of the DMA to know when a transaction has finished. 
* **Interrupt**: Interrupts will be enabled. The _window done interrupt_ is enabled if a window size is provided.
* **Interrupt wait**: The DMA HAL will block the program in a `wfi()` state until the _transaction done interrupt_ is triggered. 





### Usage
This section will explain a basic usage of the DMA as a `memcpy`, and a slightly more complex situation involving a peripheral connected via an SPI. 

#### Basic application
This example will provide a simplified code for copying data from one region of memory to another. For a real implementation please refer to the `dma_example` application in `sw/applications/dma_example/main.c`. 

The objective of this app would be to copy the content of an array into another. 

--- 

Start the application by calling 
```C
dma_init( NULL );
```

This will reset the DMA registers. The `NULL` parameter tells the HAL that the devices internal DMA is to be used. 

The most basic implementation requires the creation of **two targets** and a **transaction** relating them both. 

A target will include the information of the source or destination pointer and the characteristics of the expected transaction. 

Theoretically, there is no required difference between a _source_ and _destination_ target. They will only be differentiated once the transaction is created. Eventually, they could be interchanged from one transaction to the next. In this example they will be given explicit names and will take different arguments for the sake of clarity.

```C
static dma_target_t tgt_src = {
                                .ptr        = copy_buffer,
                                .inc_du     = 1,
                                .size_du    = sizeof(copy_buffer),
                                };
```

Here, `ptr = copy_buffer` is a `uint32_t` pointer from where the information will be extracted. `inc_du = 1` is telling the DMA that for each word copied, the pointer should be incremented by 1 unit, therefore words will be copied consecutivelly without gaps.

This configuration is implicitly initializing the rest of the target configurations as zero (as of [C99](https://gcc.gnu.org/onlinedocs/gcc/Designated-Inits.html)). This means that:
* No environment is set. 
* Data type is set to _word_ (32-bits).
* The trigger is set to _memory_ (vs. a peripheral). 

The destination target can also dispense of a size, as the source size will be used. 
  
```C
static dma_target_t tgt_dst = {
                                .ptr        = copy_buffer,
                                .inc_du     = 1,
                                };
```

Finally, a transaction is created to relate both targets:

```C
static dma_trans_t trans = {
                            .src        = &tgt_src,
                            .dst        = &tgt_dst,
                            };
```

This will also imply some configurations set to zero: 
* Mode is set to singular (only one transaction will be executed).
* There will be no window interrupts. 
* The application will have to do polling to know when the transaction has finished. 

The transaction can now be created (some extra configurations are computed from the targets), loaded into the DMA and launched.  

```C
dma_create_transaction( &trans, DMA_ENABLE_REALIGN, DMA_PERFORM_CHECKS_INTEGRITY );
dma_load_transaction( &trans );    
dma_launch( &trans );

```

When creating the transaction, the two last parameters are allowing the DMA to perform integrity checks and try some fixes (at the expense of efficiency) if misalignments are found. 

As there will be no interrupts set, the application has to check by itself the status of the transaction. 

```C
while( ! dma_is_ready() ){}
// The transaction has finished! 
```


#### Complete Application
The objective of this example is to show various special cases, precautions and considerations that can be taken. For a real example refer to the `spi_flash_write` application in `sw/applications/spi_flash_write/main.c`.

This example will copy every other byte from a buffer in RAM to a FLASH connected via an SPI to the DMA. Then, it will copy information from an SPI peripheral continuously into a small buffer. 

This explanation assumes you have read the previous example. 

---

Start the application by calling 
```C
dma_init( NULL );
```

The first source target will be pointing to a `uint16_t` buffer in memory. Data will be copied in chunks of 1 byte. For every half-word (16 bits), the only the second half (second 8 bits) will be copied.  

```C
static dma_target_t tgt1= {
                            .ptr        = (uint8_t*)(copy_buffer + 1),
                            .inc_du     = 2,
                            .size_du    = sizeof(copy_buffer),
                            .type       = DMA_DATA_TYPE_BYTE,
};

```

To start copying information from the second byte of `copy_buffer`, the source pointer is set to the address of `copy_buffer` plus one byte.
In order to copy one byte and skip another, the data type is set to byte, and the increment is set to two data units (i.e. two bytes). As for every half-word of the buffer a data unit will be copied, the transaction size is the same as the size of the `copy_buffer`. 

 
The second target will point to the address of the SPI FLASH transmission FIFO. On this example, this value was already computed and stored in a constant `ADDRESS_SPI_FLASH_TX_FIFO`. 


```C
static uint32_t *spi_flash_fifo_tx = ADDRESS_SPI_FLASH_TX_FIFO;


static dma_target_t tgt2= {
                            .inc_du     = 0,
                            .trig       = DMA_TRIG_SLOT_SPI_FLASH_TX,
};

tgt2.ptr = spi_flash_fifo_tx; 

```

Because structure initializers need to be constants, the pointer value (which is a non-constant variable) needs to be initialized outside the designated initializer. 


There is no need to assign a value of transaction size or data type. By default, the DMA will use the source's values. The data type could eventually be modified by the HAL if there is a misalignment and `DMA_ENABLE_REALIGN` is set when creating the transaction. 

The increment needs to be set to zero as the pointer should always be set to the FIFO address.

Because the SPI FLASH transmission FIFO has a line connected to slot number 4 (codified as a `1` in the fourth bit of the `trig` element) to let the DMA know if the FIFO is full, a trigger is set in that position by passing `.trig = DMA_TRIG_SLOT_SPI_FLASH_TX`.


The transaction is formed by selecting the source and destination targets

```C
static dma_trans_t trans = {
                            .src = &tgt1,
                            .dst = &tgt2,
                            .end = DMA_TRANS_END_INTR,
                            };
```

To enable interrupts, the end event is set to `DMA_TRANS_END_INTR`.


To create the transaction allowing the DMA to perform necessary realignments and integrity checks such arguments are passed to the creation function along with a pointer to the transaction.

```C
dma_config_flags_t res;
res = dma_create_transaction( &trans, DMA_ENABLE_REALIGN, DMA_PERFORM_CHECKS_INTEGRITY );
``` 

The result variable `res` contains configuration flags that can be used to check that no errors or warnings were raised. These are codified as bits in the 16-bit `res` variable. 
If only one flag was raised, the value of `res` will be equal to it. Otherwise, they can be checked by selecting individual bits, with inequalities, or comparing the result to a bitwise-or of flags. 

```C
if( res == DMA_CONFIG_OK ) // Everything was ok. 
if( res & DMA_CONFIG_CRITICAL_ERROR ) // A critical error was found. It has to be fixed or the transaction will not be loaded. 
if( res >= DMA_CONFIG_OVERLAP && res =< DMA_CONFIG_TRANS_OVERRIDE ) // One or more warning flags were raised between those two. 
if( res & ( DMA_CONFIG_SRC | DMA_CONFIG_DST) ) // Either one of the flags was raised. 
```

The transaction can then be loaded into the DMA registers by calling the load function
```C
res = dma_load_transaction( &trans );
```

and launched
```C
res = dma_load_transaction( &trans );
```

Because the DMA will raise an interrupt as soon as it has sent all of its information, it is recommended to wait for the SPI interrupt. 
```C
while( spi_intr_flag == 0 )
{
    wait_for_interrupt(); 
}
```

If something is to be done as soon as the DMA finishes (like preparing a new transaction) it can be triggered by the interrupt attention routine: 
```C
void dma_intr_handler_trans_done()
{
    // Raise a flag to trigger an action
}
```
This function is a strong implementation defined in the application of the weak one in `dma.c`. 


During this transaction
* The DMA grabbed the first byte of from the pointer (i.e. the second of the buffer, letter `B` in the chart below).
* Sent it to the SPI, where it will be trated as a 32-bit word. 
* The flash stored the byte wherever it was configured to. 
* The source pointer was increased by two data units (i.e. two bytes) to point letter `D` from the chart. 
* The DMA grabbed that byte and sent it to the SPI. 
* The process went on until the the last byte (letter `L`) is sent to the SPI.   


<table class="tg">
<thead>
  <tr>
    <td class="tg-wp8o" colspan="4">RAM</td>
    <td class="tg-wp8o" rowspan="5"></td>
    <td class="tg-wp8o" colspan="4">FLASH</td>
  </tr>
  <tr>
    <td class="tg-wp8o">A</td>
    <td class="tg-wp8o">B</td>
    <td class="tg-wp8o">C</td>
    <td class="tg-wp8o">D</td>
    <td class="tg-wp8o">0</td>
    <td class="tg-wp8o">0</td>
    <td class="tg-wp8o">0</td>
    <td class="tg-wp8o">B</td>
  </tr>
  <tr>
    <td class="tg-wp8o">E</td>
    <td class="tg-wp8o">F</td>
    <td class="tg-wp8o">G</td>
    <td class="tg-wp8o">H</td>
    <td class="tg-wp8o">0</td>
    <td class="tg-wp8o">0</td>
    <td class="tg-wp8o">0</td>
    <td class="tg-wp8o">D</td>
  </tr>
  <tr>
    <td class="tg-wp8o">I</td>
    <td class="tg-wp8o">J</td>
    <td class="tg-wp8o">K</td>
    <td class="tg-wp8o">L</td>
    <td class="tg-wp8o">0</td>
    <td class="tg-wp8o">0</td>
    <td class="tg-wp8o">0</td>
    <td class="tg-wp8o">F</td>
  </tr>
  <tr>
    <td class="tg-wp8o" colspan="4">...</td>
    <td class="tg-wp8o" colspan="4">...</td>
  </tr>
</thead>
</table>


---


For the second part of this example, the targets will be modified and their role switched. 

The source of the data will be a peripheral connected to the SPI, which will be continuously sending information in half-word format, in streams of 4096 bytes (2048 half-words). Every 1024 half-words there should be a `MILESTONE-CHARACTER`, if there is not, the data flow should be stopped. 

```C
tgt2.ptr        = spi_peripheral_fifo_rx;
tgt2.size_du    = 2048;
tgt2.type       = DMA_DATA_TYPE_HALF_WORD;   
tgt2.trig       = DMA_TRIG_SLOT_SPI_RX;
```

The source pointer is set to the reception FIFO of the SPI, and the appropriate slot is chosen. 
The increment is kept in zero. 

The destination pointer will be given by a function.

```C 
tgt1.ptr        = (uint8_t*) myApp_getDestination();
```

To guarantee that the result of this function will not cause the DMA to write in undesired regions of memory, an environment can be created. 
It requires two pointers to the first and last byte where the DMA will be authorized to make any action. 
```C 
static dma_env_t safe_zone = {
                                .start  = ADDRESS_START_OF_SAFE_ZONE,
                                .end    = ADDRESS_END_OF_SAFE_ZONE,
};
```

The environment can be assigned to the target. 
```C
tgt1.env    = &safe_zone;
tgt1.inc_du = 1;
``` 

Size and data type is up to the source, and the rest of the configurations are inherited from the last transaction.

The transaction is conformed as:
```C
window_size_du  = 1024;

trans.src       = &tgt2;
trans.dst       = &tgt1;
trans.mode      = DMA_TRANS_MODE_CIRCULAR;
trans.win_du    = window_size_du;

```

Meaning that the information will start to flow into the target 1 from target 2, and every 1024 half-words transferred the application will get an interrupt to check if the `MILESTONE_SYMBOL` is present. Upon finishing the transaction, it shall start again to refill the buffer as it was set in `CIRCULAR_MODE`.

The amount of times the buffer was filled will be updated on every _transaction done_ interrupt.
```C
void dma_intr_handler_trans_done()
{
    transaction_count++;
}
```

The search for the `MILESTONE_SYMBOL` can be done inside the _window done_ interrupt handling.
```C
void dma_intr_handler_window_done()
{
    /* The current window is obtained. The count is zero when no windows have yet been written. When it is set to one, the window zero is ready. */
    window_count    = dma_get_window_count() -1;
    /* The pointer to the symbol is obtained from the destination pointer + the amount of half-words that have been written. this assumes the symbol is on the first element of each chunk.*/   
    address         = (uint16_t *)trans.dst->ptr + window_count*window_size_du;
    symbol          = *address;
    if( symbol != MILESTONE_SYMBOL )
    {
        /* If the symbol was not the expected one, future transactions should not be carried out.*/
        dma_stop_circular();
        /* The number of the first window with error is saved to analyze it later.*/
        error_window = error_window == 0 ? window_count : error_window;
    }
}
```
> :warning: Interrupt attention routines are adviced to be kept as short as possible. This example is merely illustrative. 

Once the `dma_stop_circular()` function is called, the DMA will still finish the transaction it is currently executing (which could be a whole transaction if the faulty window was the last one). It is important to save valuable information that might be overriden by the ongoing transaction (like the faulty window number). 


------
DMA HAL-Updated Hardware

JuanSapriza:
I have finished the documentation. It is supposed to complement the one present in the code. 
Please add any comments/questions you might have so it can be improved. I would like that anyone could read it and be ready to start coding a white test for the DMA.

JuanSapriza:
In the process of rebasing to compatibilize with #203

JuanSapriza:
@fimtrey Could you try testing the I2S application? I dont have the hardware to do so. 

Also, we are slowly drifting towards apps that can be tested (to make sure that future changes do not break them). Do you think of any way that we could test this app without requiring the microphone? 
Maybe, for simulation, we could fake the answer of the microphone. 
Any ideas @JoseCalero @davideschiavone ?

JuanSapriza:
[Here is a spreadsheet](https://docs.google.com/spreadsheets/d/1hXeQu6oHp9lK5JQJ8nYAGD4jo1BzW_Ddz3hJdvgP0zU/edit?usp=sharing) that we can use to analyze all the pathological cases and what is the expected return of the HAL. 
It is incomplete, but its a draft that we could use.

fimtrey:
> @fimtrey Could you try testing the I2S application? I don't have the hardware to do so.

I do not have the hardware anymore...
If you have the FPGA ready, I may be able to drop by and plug in the microphone. 
 
> Also, we are slowly drifting towards apps that can be tested (to make sure that future changes do not break them). Do you think of any way that we could test this app without requiring the microphone? I noticed that this application needs `MEMORY_BANKS=16`, I guess due to the size of the data buffer which is `#define AUDIO_DATA_NUM 0x18000`. I ran it with just 100 words and 16 batches to test it (so it actually ends some day). Would you say this is a viable configuration to leave as default so we can use to test?

This is a demo example so: if somebody wants to run it, fewer than 8 MEM_BANKS are not really interesting, as you get less than 1 second of audio. I had 2048 but @davideschiavone said to put the max. You can put whatever.

It's stupid to store such big amount of data in RAM but there is no second DMA channel to transfer it to the FLASH... in future there will be better ways and no need for 16 MEM_BANKS. 

For testing I'd stick to the simulation.

> Maybe, for simulation, we could fake the answer of the microphone. Any ideas @JoseCalero @davideschiavone ?

This exist, if you build verilator with the external flag (see [ExternalDevices.md](https://github.com/esl-epfl/x-heep/blob/main/ExternalDevices.md))
```
make verilator-sim FUSESOC_FLAGS="--flag=use_external_device_example"
```

JuanSapriza:
@JoseCalero I could not mark your requested changes as done. Could you take a look. Instead I re-requested your review (hopefully that resets the fla ^^ )

JuanSapriza:
> > @fimtrey Could you try testing the I2S application? I don't have the hardware to do so.
> 
> I do not have the hardware anymore... If you have the FPGA ready, I may be able to drop by and plug in the microphone.
> 
> > Also, we are slowly drifting towards apps that can be tested (to make sure that future changes do not break them). Do you think of any way that we could test this app without requiring the microphone? I noticed that this application needs `MEMORY_BANKS=16`, I guess due to the size of the data buffer which is `#define AUDIO_DATA_NUM 0x18000`. I ran it with just 100 words and 16 batches to test it (so it actually ends some day). Would you say this is a viable configuration to leave as default so we can use to test?
> 
> This is a demo example so: if somebody wants to run it, fewer than 8 MEM_BANKS are not really interesting, as you get less than 1 second of audio. I had 2048 but @davideschiavone said to put the max. You can put whatever.
> 
> It's stupid to store such big amount of data in RAM but there is no second DMA channel to transfer it to the FLASH... in future there will be better ways and no need for 16 MEM_BANKS.
> 
> For testing I'd stick to the simulation.
> 
> > Maybe, for simulation, we could fake the answer of the microphone. Any ideas @JoseCalero @davideschiavone ?
> 
> This exist, if you build verilator with the external flag (see [ExternalDevices.md](https://github.com/esl-epfl/x-heep/blob/main/ExternalDevices.md))
> 
> ```
> make verilator-sim FUSESOC_FLAGS="--flag=use_external_device_example"
> ```

Great! Thanks @fimtrey, I will test it and leave a note that it can only be run under those circumstances.

JuanSapriza:
I will close this PR as there are many differences being shown in the _changed files_ window. The [new PR](https://github.com/esl-epfl/x-heep/pull/317) does not have this issue.

Issue #226, opened 2023-03-16
Tie XIF to 0 in x_heep_system if X_EXT=0


davidmallasen:
This would remove the need of creating a dummy xif and connecting it to the ports of `x_heep_system` as done in the `tb/testharness.sv` with:
~~~
// eXtension Interface
  if_xif #() ext_if ();

...
      .xif_compressed_if(ext_if),
      .xif_issue_if(ext_if),
      .xif_commit_if(ext_if),
      .xif_mem_if(ext_if),
      .xif_mem_result_if(ext_if),
      .xif_result_if(ext_if),
...
~~~

davideschiavone:
is this done @davidmallasen ?

davidmallasen:
I don't think so, it depends on what we want to do. Now we leave it unconnected in the testharness and in the cheeps we do the same and tie it to 0

PR #233, opened 2023-03-22
OpenRoad Flow

BaranErdo:


davidmallasen:
Why remove the dependency on `openhwgroup.org:ip:cv32e40x` ? Also, from what I see this is a workaround to issue #226 . Could we fix both things at the same time? I believe it makes more sense to use the X_EXT parameter instead of the YOSYS defines.

PR #235, opened 2023-03-28
updating fast intr ctrl based on structure

Ho3eintaji:
### changes
- Writing based on the format defined for the whole SW layer
- Changing to use the new structure for reading/writing from peripherals
- Removing unnecessary variables like base address, considering the fact that now registers structure is memory aligned
- Now ***clear_fast_interrupt*** requires only one parameter (only fast intr type) rather than two in the previous version. The second parameter was basically passed for the base address which is not required anymore.
```c
fast_intr_ctrl_result_t clear_fast_interrupt(fast_intr_ctrl_fast_interrupt_t fast_interrupt)
```


sw/gdbInit
JoseCalero, line 1: this should not be included

Ho3eintaji: So you mean adding it to gitignore?!

Ho3eintaji:
> Please implement in the FIC HAL the (weak) handlers of the irqs connected to it, check the handler.c, and also add the clearing bit to those functions

Ho3eintaji:
> > Please implement in the FIC HAL the (weak) handlers of the irqs connected to it, check the handler.c, and also add the clearing bit to those functions
> 
> Ok I think, there should be a header (whatever its name) providing all the weak functions (wherever they are from PLIC or FIC) for the developer because I think there is no point in separating them for the developer. For clearing the bit, doesn't it limit the developer, they should decide when removing this bit?!

So you mean the structure should be like this for FIC:
an strong function which clean the rcv intr > then proving weak for all the FIC peripherals > then inside peripherals (SPI, DMA,..) they can provide another level, and weak function they provide is the one, developer is touching?!

JoseCalero:
> > Please implement in the FIC HAL the (weak) handlers of the irqs connected to it, check the handler.c, and also add the clearing bit to those functions
> 
> Ok I think, there should be a header (whatever its name) providing all the weak functions (wherever they are from PLIC or FIC) for the developer because I think there is no point in separating them for the developer. For clearing the bit, doesn't it limit the developer, they should decide when removing this bit?!

The PLIC works in a different way than the FIC. In fact, the PLIC generates 1 irq to the CPU, and then you check the source by reading the PLIC registers. In the FIC, you generate as many irqs as connected sources, this is why you have in the vector.S different "_fast" entries. Then, in the handler.c file you have all the different irqs, but this file is within `runtime` folder, which is a folder for cpu-oriented stuff. Thus, if we want to keep coherence and also provide self-contained modules, those weak implementations should be contained within the FIC HAL. Otherwise, the developer has to remember that in case he/she has different connections into the FIC, he has to go to `runtime\handler.c` and modify there whatever is needed. It is not a major problem, but I think we can provide a more logical sense by having the HAL things in their respective HAL modules.
And the thing about clearing the bit, is just to put something in the weak implementation (instead of the while(1) that right now is employed..).
I let @StefanoAlbini96, @JuanSapriza, and @ruben-roalvarez comment on this as well

Ho3eintaji:
> > > Please implement in the FIC HAL the (weak) handlers of the irqs connected to it, check the handler.c, and also add the clearing bit to those functions
> > 
> > 
> > Ok I think, there should be a header (whatever its name) providing all the weak functions (wherever they are from PLIC or FIC) for the developer because I think there is no point in separating them for the developer. For clearing the bit, doesn't it limit the developer, they should decide when removing this bit?!
> 
> The PLIC works in a different way than the FIC. In fact, the PLIC generates 1 irq to the CPU, and then you check the source by reading the PLIC registers. In the FIC, you generate as many irqs as connected sources, this is why you have in the vector.S different "_fast" entries. Then, in the handler.c file you have all the different irqs, but this file is within `runtime` folder, which is a folder for cpu-oriented stuff. Thus, if we want to keep coherence and also provide self-contained modules, those weak implementations should be contained within the FIC HAL. Otherwise, the developer has to remember that in case he/she has different connections into the FIC, he has to go to `runtime\handler.c` and modify there whatever is needed. It is not a major problem, but I think we can provide a more logical sense by having the HAL things in their respective HAL modules. And the thing about clearing the bit, is just to put something in the weak implementation (instead of the while(1) that right now is employed..). I let @StefanoAlbini96, @JuanSapriza, and @ruben-roalvarez comment on this as well

Ok, make sense. I will add the weak handlers inside FIC. However, its better for a developer to have access to each peripheral irq handler (either it comes from FIC or PLIC) inside its driver. So drivers written for peripherals (dma, spi, ...) use FIC provided weak handlers. The only thing that should be decided is to clean the intr bit inside peripheral drivers or FIC.

sw/device/lib/drivers/fast_intr_ctrl/fast_intr_ctrl.c
```diff
@@ -1,24 +1,431 @@
...
-    mmio_region_write32(fast_intr_ctrl->base_addr, (ptrdiff_t)(FAST_INTR_CTRL_FAST_INTR_CLEAR_REG_OFFSET), reg);
-
+    /* write in reg through structure */
+    fast_intr_ctrl_peri->FAST_INTR_CLEAR = reg;
     return kFastIntrCtrlOk_e;
 }
+
+__attribute__((weak)) void fic_irq_timer_1(void)
+{
+    volatile uint8_t i;
```
JoseCalero, line 227: remove this, add:
`/* Users should implement their non-weak version */`

sw/device/lib/drivers/fast_intr_ctrl/fast_intr_ctrl.c
```diff
@@ -1,24 +1,431 @@
...
 #include "mmio.h"
-
 #include "core_v_mini_mcu.h"
-
 #include "fast_intr_ctrl_regs.h"  // Generated.
+#include "fast_intr_ctrl_structs.h"
+
+/* interrupts */
+// #include "handler.h"
+#include "csr.h"
```
JoseCalero, line 43: why do you need to include this?

sw/device/lib/drivers/fast_intr_ctrl/fast_intr_ctrl.h
```diff
@@ -1,60 +1,236 @@
...
+/**                          INLINE FUNCTIONS                              **/
+/**                                                                        **/
+/****************************************************************************/
 
+/****************************************************************************/
+/**                                                                        **/
+/**                                EOF                                     **/
+/**                                                                        **/
+/****************************************************************************/
 #ifdef __cplusplus
```
JoseCalero, line 234: remove this, this is c-based header

sw/device/lib/drivers/fast_intr_ctrl/fast_intr_ctrl.h
```diff
@@ -1,60 +1,236 @@
...
+/**
+* @file   fast_intr_ctrl.h
+* @date   27/03/23
+* @brief  The fast interrupt controller peripheral driver 
+*/
 
 #ifndef _FAST_INTR_CTRL_H_
 #define _FAST_INTR_CTRL_H_
 
+#ifdef __cplusplus
```
JoseCalero, line 30: remove this, this is c-based header

JoseCalero (changes requested):
almost ready :)

sw/device/lib/drivers/fast_intr_ctrl/fast_intr_ctrl.c
Ho3eintaji: It is not working in this way, if I replace it strong function never calls

sw/device/lib/drivers/fast_intr_ctrl/fast_intr_ctrl.c
Ho3eintaji: you are right, removed, it was there from pre version

sw/device/lib/drivers/fast_intr_ctrl/fast_intr_ctrl.c
```diff
@@ -1,24 +1,424 @@
...
+/**                                                                        **/
+/*                             MODULES USED                                 */
+/**                                                                        **/
+/****************************************************************************/
 
 #include "fast_intr_ctrl.h"
+#include "core_v_mini_mcu.h"
+#include "fast_intr_ctrl_regs.h"  // Generated.
+#include "fast_intr_ctrl_structs.h"
+// #include "csr.h"
```
JoseCalero, line 37: remove that comment if such include is not used

sw/device/lib/drivers/fast_intr_ctrl/fast_intr_ctrl.c
JoseCalero: yes, you have to change the 
`__attribute__((weak))`
to
`__attribute__((weak, optimize("O0")))`
Note that we are compiling the whole thing with '0s', i.e. all optimisations till '02' are active. If we specify just '__weak__' dead-code optimisation is applied over it

JoseCalero (changes requested):
Last change

sw/device/lib/drivers/fast_intr_ctrl/fast_intr_ctrl.c
Ho3eintaji: DONE

Ho3eintaji:
> Last change

applied

JuanSapriza:
Hello @JoseCalero @Ho3eintaji One question about this implementation. 
I noticed that the handler functions have this structure: 
```C
void handler_irq_fast_xxx(void)
{
    // The interrupt is cleared.
    clear_fast_interrupt(kxxx_fic_e);
    // call the weak fic handler
    fic_irq_xxx();
}
```

Couldn't it be an issue that we are clearing the interrupt and only afterwards calling the weak implementation? Shouldn't we do it the other way around? 
I guess it could happen that we receive an interrupt and we enable the triggering of the following one before we have attended the first. We had an issue of this kind in the DMA when we have several interrupts, one after the other. If they are not blocked we end up in a loop inside the handler.

Untitled drawing(1)

JoseCalero:
It depends on the priority of intr3

Ho3eintaji:
@JuanSapriza, @JoseCalero  yeah, for sure, FIR clears the bit when it receives an intr. We did it for the sake of simplifying the process for higher layers.
However, if there is an application that it prefers to be blocked until isr being finished (in case it may miss interrupts while the bit is not cleared), we can change it to give the ability to clear the bit in the application.
If it was me, I preferred this structure to keep ISR short to prevent this. 
However, if what you said is preferred for dma, we can change it

PR #240, opened 2023-03-31
rv_plic HAL refactor

StefanoAlbini96:
Changed the HAL of the rv_plic according to the automatically-generated struct containing the info of the registers.


sw/device/lib/drivers/rv_plic/rv_plic.c
```diff
@@ -1,8 +1,44 @@
...
+* @brief  This is the main file for the HAL of the RV_PLIC peripheral
+*
+* In this file there are the defintions of the HAL functions for the RV_PLIC
+* peripheral. They provide many low level functionalities to interact
+* with the registers content, reading and writing them according to the specific
+* function of each one.
+*
+*/
+
+#define _RV_PLIC_C_SRC
```
JoseCalero, line 32: if you are not using this in the header, you can remove this

sw/device/lib/drivers/rv_plic/rv_plic.h
```diff
@@ -135,121 +158,171 @@ typedef enum dif_plic_irq_trigger {
...
-dif_plic_result_t dif_plic_irq_is_pending(const dif_plic_t *plic,
-                                          dif_plic_irq_id_t irq,
-                                          bool *is_pending);
+ * Flag used to handle the wait for interrupt.
+ * The core can test this variable to check if it has to wait
+ * for an interrupt coming from the RV_PLIC.
+ * When an interrupt occurs, this flag is set to 1 by the plic in order
+ * for the core to continue with the execution of the program.
+*/
+extern int8_t external_intr_flag;
```
JoseCalero, line 191: i do not understand why this is here: 1) it is under global section in C file, 2) this variable should be on the user-scope and not on the HAL

StefanoAlbini96: I use it inside rv_plic.c so that the handler function can change it after calling the handler for the specific interrupt source. 
I also need it in the rv_plic.h so I can access it from the main (by including the header) and use it as a wile-loop variable when I call the wait for interrupt.

```
  while(external_intr_flag==0) {
      gpio_write(&gpio, GPIO_TB_OUT, true);
      wait_for_interrupt();
  }
```

sw/device/lib/drivers/rv_plic/rv_plic.c
```diff
@@ -12,20 +46,33 @@
 #include "mmio.h"
 
 #include "rv_plic_regs.h"  // Generated.
+#include "core_v_mini_mcu.h"
```
JoseCalero, line 49: you are actually adding this twice, as it is already added into rv_plic_structc.h, right?

sw/device/lib/drivers/rv_plic/rv_plic.c
```diff
@@ -34,337 +81,417 @@ typedef struct plic_reg_info {
...
+/**                                                                        **/
+/****************************************************************************/
+
+/****************************************************************************/
+/**                                                                        **/
+/*                            GLOBAL VARIABLES                              */
+/**                                                                        **/
+/****************************************************************************/
+
+int8_t external_intr_flag = 0;
```
JoseCalero, line 108: why is it signed? can you put a comment to specify the context of this variable (what is its purpose etc?)?

sw/device/lib/drivers/rv_plic/rv_plic.c
```diff
@@ -34,337 +81,417 @@ typedef struct plic_reg_info {
...
+
+/****************************************************************************/
+/**                                                                        **/
+/*                           EXPORTED FUNCTIONS                             */
+/**                                                                        **/
+/****************************************************************************/
+
+__attribute__((weak)) void handler_irq_uart(void) {
```
JoseCalero, line 116: please, avoid the printing etc, and add as it was done in the FIC

sw/device/lib/drivers/rv_plic/rv_plic.c
```diff
@@ -34,337 +81,417 @@ typedef struct plic_reg_info {
...
-static plic_reg_info_t plic_irq_enable_reg_info(dif_plic_irq_id_t irq,
-                                                dif_plic_target_t target) {
-  ptrdiff_t offset = plic_offset_from_reg0(irq);
-  return (plic_reg_info_t){
-      .offset = plic_irq_enable_base_for_target(target) + offset,
-      .bit_index = plic_irq_bit_index(irq),
-  };
+ * Returns the irq_sources_t source type of a given irq source ID
+*/
+static irq_sources_t plic_get_irq_src_type(plic_irq_id_t irq_id)
```
JoseCalero, line 144: if this is static, then why is it here? where is its declaration?

sw/device/lib/drivers/rv_plic/rv_plic.c
```diff
@@ -34,337 +81,417 @@ typedef struct plic_reg_info {
...
+{
+  if(irq_id >= UART_ID_START && irq_id <= UART_ID_END)
```
JoseCalero, line 146: Try to put always negative-first implementations in multi-branch code, by doing so you can avoid one comparison per if

sw/device/lib/drivers/rv_plic/rv_plic.c
```diff
@@ -34,337 +81,417 @@ typedef struct plic_reg_info {
...
-static plic_reg_info_t plic_irq_pending_reg_info(dif_plic_irq_id_t irq) {
-  ptrdiff_t offset = plic_offset_from_reg0(irq);
-  return (plic_reg_info_t){
-      .offset = RV_PLIC_IP_0_REG_OFFSET + offset,
-      .bit_index = plic_irq_bit_index(irq),
-  };
-}
+void handler_irq_external(void)
+{
+  plic_irq_id_t int_id;
```
JoseCalero, line 169: initialize everything

sw/device/lib/drivers/rv_plic/rv_plic.c
```diff
@@ -34,337 +81,417 @@ typedef struct plic_reg_info {
...
+  plic_result_t res = plic_irq_claim(0, &int_id);
```
JoseCalero, line 170: avoid magic numbers

sw/device/lib/drivers/rv_plic/rv_plic.c
JoseCalero: also, i think we can optimize this by using function listing

sw/device/lib/drivers/rv_plic/rv_plic.c
```diff
@@ -34,337 +81,417 @@ typedef struct plic_reg_info {
...
+  irq_sources_t type = plic_get_irq_src_type(int_id);
+
+  switch (type)
```
JoseCalero, line 173: i think we can optimize this by using function listing, no need for the switch

sw/device/lib/drivers/rv_plic/rv_plic.c
```diff
@@ -34,337 +81,417 @@ typedef struct plic_reg_info {
...
+  Resets relevant registers of the PLIC (Level/Edge, 
+  priority, target, threshold, interrupts).
+  It writes the default value 0 in them and read back
+  the value into the proper register struct
+*/
+plic_result_t plic_Init(void)
+{
+  
+  /* Clears all the Level/Edge registers */ 
+  for(size_t i=0; i<RV_PLIC_LE_MULTIREG_COUNT; i++)
```
JoseCalero, line 217: try to avoid using size_t, it does not give a correct sense on how big is RV_PLIC_LE_MULTIREG_COUNT in this case

sw/device/lib/drivers/rv_plic/rv_plic.c
```diff
@@ -34,337 +81,417 @@ typedef struct plic_reg_info {
...
+  {
+    uint32_t *dest = (&rv_plic_peri->IP0 + i);
+    *dest = 0;
+
+    if (*dest != 0)
```
JoseCalero, line 222: MINOR: we need to discuss this, basically, you are assuming that there could be a problem where the register cannot be set, right? I mean... can this really happen?

sw/device/lib/drivers/rv_plic/rv_plic.c
```diff
@@ -34,337 +81,417 @@ typedef struct plic_reg_info {
...
-      is_pending == NULL) {
-    return kDifPlicBadArg;
-  }
-
-  ptrdiff_t msip_offset = plic_software_irq_base_for_target(target);
-  uint32_t register_value =
-      mmio_region_read32(plic->params.base_addr, msip_offset);
 
-  *is_pending = (register_value == 1) ? true : false;
+plic_result_t plic_software_irq_is_pending(bool *is_pending)
```
JoseCalero, line 468: do not waste one reference pass + return variable, just use the return variable to say true or false

JoseCalero (changes requested):
First review: changes requested

JoseCalero:
Please, update this PR with the last changes in main and so to avoid conflicts etc

PR #246, opened 2023-04-05
update GPIO driver

Ho3eintaji:
- Full GPIO driver
- Driver functions are put in several levels of abstraction. Although abstract functions like gpio_config wrap up several function calls by passing a configuration structure, the driver still provides lower-level functions.
- Apps are updated based on the driver


davideschiavone:
please check all the applications:)

Ho3eintaji:
> please check all the applications:)

All the affected apps were updated

sw/applications/blinky_freertos/main.c
```diff
@@ -251,25 +248,22 @@ void system_init(void)
...
+    pin_cfg.pin = GPIO_LD5_B;
+    gpio_res |= gpio_config(pin_cfg);
+    pin_cfg.pin = GPIO_LD5_G;
+	gpio_res |= gpio_config(pin_cfg);
     if (gpio_res != kGpioOk) printf("Failed\n;");
-	gpio_write(&gpio, GPIO_LD5_G, false);
+    
+    gpio_write(GPIO_LD5_R, false);
+    gpio_write(GPIO_LD5_B, false);
+	gpio_write(GPIO_LD5_G, false);
```
davideschiavone, line 266: pls  careful with the identation

sw/applications/example_power_gating_core/main.c
```diff
@@ -22,13 +22,13 @@ static const uint64_t kTickFreqHz = 1000 * 1000; // 1 MHz
...
 static dif_plic_t rv_plic;
-static gpio_t gpio;
+// static gpio_t gpio;
 
 int main(int argc, char *argv[])
 {
     // Setup fast interrupt controller
-    fast_intr_ctrl_t fast_intr_ctrl;
-    fast_intr_ctrl.base_addr = mmio_region_from_addr((uintptr_t)FAST_INTR_CTRL_START_ADDRESS);
+    // fast_intr_ctrl_t fast_intr_ctrl;
```
davideschiavone, line 30: remove commented lines - why are they kept?

sw/applications/example_power_gating_core/main.c
```diff
@@ -56,9 +56,9 @@ int main(int argc, char *argv[])
     rv_timer_init(timer_2_3_reg, (rv_timer_config_t){.hart_count = 2, .comparator_count = 1}, &timer_2_3);
 
     // Setup gpio
-    gpio_params_t gpio_params;
-    gpio_params.base_addr = mmio_region_from_addr((uintptr_t)GPIO_START_ADDRESS);
-    gpio_init(gpio_params, &gpio);
+    // gpio_params_t gpio_params;
```
davideschiavone, line 59: same here and all the others

davideschiavone (changes requested):

sw/applications/example_power_gating_core/main.c
Ho3eintaji: I just forgot, now removed

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,624 @@
...
+    {
+        return GpioError;
+    }
+}
+
+gpio_result_t gpio_intr_check_stat_rise (gpio_pin_number_t pin, bool *is_pending)
+{
+    if (pin >= 0 && pin < MAX_PIN)
+    {
+        if (getBitfield(gpio_peri->INTRPT_RISE_STATUS0, 1, pin) == 1)
```
JoseCalero, line 459: 'getBitfield(gpio_peri->INTRPT_RISE_STATUS0, 1, pin)' don't we have enums or defines to avoid magic numbers?

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,624 @@
...
+        }
+        return GpioOk;
+    }
+    else
+    {
+        return GpioError;
+    }
+}
+
+gpio_result_t gpio_intr_check_stat_rise (gpio_pin_number_t pin, bool *is_pending)
```
JoseCalero, line 455: I suggest avoiding passing reference to bool. Instead, the return value could be used to codify that, at the end you are using the return value to check almost as that

JoseCalero: if there is a need to codify more gpio_result_t enums, do it

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,624 @@
...
+    {
+        return GpioError;
+    }
+}
 
-  return kGpioOk;
+gpio_result_t gpio_intr_set_mode (bool gpio_intr_mode)
+{
+    if (gpio_intr_mode)
+        setBitfield(&(gpio_peri->CFG), 1, 1, 0);
```
JoseCalero, line 608: can we avoid magic numbers?

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,624 @@
...
+    else
+        setBitfield(&(gpio_peri->CFG), 0, 1, 0);
+    return GpioOk;
```
JoseCalero, line 611: here, we are returning always the same regardless of the branch condition, thus, is it truly needed to return anything?

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,624 @@
...
+    else
+    {
+        return GpioError;
+    }
+}
+
+gpio_result_t gpio_intr_dis_all (gpio_pin_number_t pin){
+    if (pin >= 0 && pin < MAX_PIN)
+    {
+        setBitfield(&(gpio_peri->INTRPT_RISE_EN0),     0, 1, pin);
```
JoseCalero, line 392: can we avoid magic numbers?

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,624 @@
...
+}
+
+gpio_result_t gpio_reset (gpio_pin_number_t pin)
+{
+    if (pin >= 0 && pin < MAX_PIN)
+    {
+        gpio_intr_set_mode (0);
+        gpio_set_mode (pin, GpioModeIn);
+        gpio_dis_input_sampling (pin);
+        setBitfield(&(gpio_peri->GPIO_CLEAR0), 0, 0b1, pin);
```
JoseCalero, line 182: magic numbers

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,624 @@
...
+    /* if interrupt is enabled. Also after enabling check for error */
+    if (cfg.en_intr == true)
+        if (gpio_intr_en (cfg.pin, cfg.intr_type) == GpioError)
+            return GpioError;
+    return GpioOk;
+}
+
+gpio_result_t gpio_set_mode (gpio_pin_number_t pin, gpio_mode_t mode)
+{
+    if (pin >= 0 && pin < 16)
```
JoseCalero, line 133: magic numbers

sw/device/lib/drivers/gpio/gpio.h
```diff
@@ -1,282 +1,369 @@
...
+
+/**
+ * The structure should be used by user for configuring gpio.
+ * it includes pin number, mode, enable sampling, enable intr, and its mode.
+ */
+typedef struct gpio_cfg
+{
+    gpio_pin_number_t pin;      /*!< pin number. */
+    gpio_pin_number_t mode;     /*!< pin mode. */
+    bool en_input_sampling;     /*!< enable sampling (being input is req). */
```
JoseCalero, line 104: can you use uint8_t instead?

sw/device/lib/drivers/gpio/gpio.h
```diff
@@ -1,282 +1,369 @@
...
+ * (0) the corresponding GPIO will not sample the inputs (saves power) and 
+ * will not generate any interrupts.
+ */
+gpio_result_t gpio_en_input_sampling (gpio_pin_number_t pin);
+
+/**
+ * @brief disable sampling as input by writing zero in GPIO_EN.
+ */
+gpio_result_t gpio_dis_input_sampling (gpio_pin_number_t pin);
+
```
JoseCalero, line 169: leave one space between declarations

sw/device/lib/drivers/gpio/gpio.h
```diff
@@ -1,282 +1,369 @@
...
+/**
+ * @brief enable the interrupt for the given pin. Type of interrupt 
+ * specified by user.
+ * @param gpio_pin_number_t specify pin number
+ * @param gpio_intr_type_t specify the type of the interrupt
+ */
+gpio_result_t gpio_intr_en (gpio_pin_number_t pin, gpio_intr_type_t type);
+
+// /**
+//  * @brief disable the given type of interrupt on the given pin.
```
JoseCalero, line 264: why do we have different functions that are commented?

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,624 @@
...
+/****************************************************************************/
+/**                                                                        **/
+/*                             MODULES USED                                 */
+/**                                                                        **/
+/****************************************************************************/
 
 #include "gpio.h"
-
 #include "gpio_regs.h"  // Generated.
+#include "gpio_structs.h" //todo: remove it
```
JoseCalero, line 35: check this 'todo'

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,624 @@
...
+/**                                                                        **/
+/****************************************************************************/
+
+gpio_result_t gpio_config (gpio_cfg_t cfg){
+    // /* check that passed arg is not empty*/
+    // if (cfg == NULL)
+    //     return GpioError;
+    /* check that pin is in acceptable range. */
+    if (cfg.pin > (MAX_PIN-1) || cfg.pin < 0)
+        return GpioError;
```
JoseCalero, line 113: in this function, there are different conditions that can give an error, so the user should know what is exactly the thing that failed. Thus, I recommend implementing some error codes besides Ok and Error to help tackling/knowing this

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,624 @@
...
+//         gpio_peri->GPIO_OUT0 &= ~(1 << pin);
+//         return GpioOk;
+//     }
+//     else
+//     {
+//         return GpioError;
+//     } 
+// }
+
+//todo: check to see its toggling or just writing one
```
JoseCalero, line 254: is this 'todo' relevant? did you check it then?

sw/device/lib/drivers/gpio/gpio.h
```diff
@@ -1,282 +1,369 @@
...
+} gpio_intr_type_t;
+
+/**
+ * The structure should be used by user for configuring gpio.
+ * it includes pin number, mode, enable sampling, enable intr, and its mode.
+ */
+typedef struct gpio_cfg
+{
+    gpio_pin_number_t pin;      /*!< pin number. */
+    gpio_pin_number_t mode;     /*!< pin mode. */
```
JoseCalero, line 103: can we define an enum with the different modes?

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,624 @@
...
+    }
+    else
+    {
+        return GpioError;
+    }
+}
+
+gpio_result_t gpio_en_input_sampling (gpio_pin_number_t pin)
+{
+    if (pin >= 0 && pin < MAX_PIN)
```
JoseCalero, line 151: can we set this also for output pins?? if not, we should add some sanity check?

JoseCalero (changes requested):

sw/device/lib/drivers/gpio/gpio.h
Ho3eintaji: It is already defined, I used gpio_pin_number_t than gpio_mode_t by mistake for second field

sw/device/lib/drivers/gpio/gpio.c
Ho3eintaji: you are right, we can do the sanity check to not enable it by mistake for output pins. When it calls through gpio_config sanity check is done inside the function, however, when it is called separately it may happen, so yes good to add!

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,624 @@
...
+            *val = false;
+        return GpioOk;
+    }
+    else
+    {
+        return GpioError;
+    }
+}
+
+// gpio_result_t gpio_pin_set (gpio_pin_number_t pin)
```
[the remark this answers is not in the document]
Ho3eintaji: I decided not to use the hw registers they put for masking and then clearing and setting.
Instead, like all other registers handling bitwise masking in sw, simply writing to GPIO_OUT.
Moreover, with availability of function gpio_write, gpio_set and gpio_clear are unnecessary. However, if you disagree, I will add them too.

sw/device/lib/drivers/gpio/gpio.c
Ho3eintaji: Actually, I prefer it this way, one gpio_result for the sanity check. I will add to it if other situations should be considered. But the values should pass to a user give it back through pointers that are passed. what do you think?!

sw/device/lib/drivers/gpio/gpio.h
Ho3eintaji: removed.

sw/device/lib/drivers/gpio/gpio.c
Ho3eintaji: Ok, the checkings I was thinking to do was "pin number" being correct and also a structure in the form gpio_config being passed. Some of the checking I'm not doing internally, for example if the gpio is set as output, enable sampling is never set for it.
However, if you think those two are not enough, let me know

Ho3eintaji: I added several checkings, and based on them, the error type is different:
- Given pin is not inside of acceptable range
- Given pin mode is not one of the defined modes in gpio_mode_t
- Given interrupt type is not one of the defined types in gpio_intr_type_t

Ho3eintaji: Anything to be returned than 0 (GpioOk) is an error.

Ho3eintaji:
@JoseCalero conflicts with rv_plic commit must be resolved now.

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,525 @@
...
+#define GPIO_EN__ENABLED     1
+#define GPIO_EN__DISABLED    0
+
+/**
+ * GPIO_CLEAR and GPIO_SET putting and removing mask
+ */
+#define GPIO_CLEAR__PUT_MASK 1
+#define GPIO_CLEAR__REMOVE_MASK 0
+#define GPIO_SET__PUT_MASK 1
+#define GPIO_SET__REMOVE_MASK 0
```
JoseCalero, line 63: do not duplicate defines, just:
`#define GPIO_PUT_MASK 0x01
#define GPIO_REMOVE_MASK 0x00`

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,525 @@
...
 
 /**
- * Gives the mask that corresponds to the given bit index.
- *
- * @param index Bit index in a 32-bit register.
+ * Mask for different bit field width
  */
-static uint32_t index_to_mask(uint32_t index) { return 1u << index; }
+#define MASK_1b     0b1
+#define MASK_2b     0b11
```
JoseCalero, line 49: for these two, use the bitmasks provided in bitfield.h

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,525 @@
...
+    /* if interrupt is enabled. Also after enabling check for error */
+    if (cfg.en_intr == true)
+        if (gpio_intr_en (cfg.pin, cfg.intr_type) != GpioOk)
+            return GpioIntrTypeNotAcceptable;
+    return GpioOk;
+}
+
+gpio_result_t gpio_set_mode (gpio_pin_number_t pin, gpio_mode_t mode)
+{
+    if (pin >= 0 && pin < 16)
```
JoseCalero, line 149: can we add here a define based on the max number of pins? (imagine that in the future they change...)

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,525 @@
...
+        *val = false;
+    return GpioOk;
+}
 
-  if (gpio == NULL) {
-    return kGpioBadArg;
-  }
+//todo: check it
+gpio_result_t gpio_toggle (gpio_pin_number_t pin)
+{
```
JoseCalero, line 230: why are you not using the `GPIO_TOGGLE0` register?

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,525 @@
...
+}
 
-  uint32_t reg_value = mmio_region_read32(gpio->params.base_addr, GPIO_GPIO_OUT_REG_OFFSET);
-  reg_value = bitfield_bit32_write(reg_value, pin, state);
-  mmio_region_write32(gpio->params.base_addr, GPIO_GPIO_OUT_REG_OFFSET, reg_value);
+gpio_result_t gpio_write (gpio_pin_number_t pin, bool val)
+{
+    if (pin > (MAX_PIN-1) || pin < 0)
+        return GpioPinNotAcceptable;
+    if (val == GPIO_IS_SET)
```
JoseCalero, line 250: you can avoid the if else, use the val argument, a bool is expanded to integer (1 for true, 0 for false)

sw/device/lib/drivers/gpio/gpio.c
JoseCalero: you can let the pointer as a reference if you want, but you do not need the if else statement, you can use the direct output of the read operation to set the `is_pending`

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,525 @@
...
+
+gpio_result_t gpio_intr_check_stat_lvl_low (gpio_pin_number_t pin, bool *is_pending)
+{
+    if (pin > (MAX_PIN-1) || pin < 0)
+    {   
+        *is_pending = GPIO_INTR_IS_NOT_TRIGGERED;
+        return GpioPinNotAcceptable;
+    }
+        
+    if (bitfield_read(gpio_peri->INTRPT_LVL_LOW_STATUS0, MASK_1b, pin) ==
```
JoseCalero, line 423: here the same, you can use the result from the read operation

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,525 @@
...
+gpio_result_t gpio_intr_check_stat_fall (gpio_pin_number_t pin, bool *is_pending)
+{
+    if (pin > (MAX_PIN-1) || pin < 0)
+    {   
+        *is_pending = GPIO_INTR_IS_NOT_TRIGGERED;
+        return GpioPinNotAcceptable;
+    }
+        
+    if (bitfield_read(gpio_peri->INTRPT_FALL_STATUS0, MASK_1b, pin) == 
+        GPIO_INTR_IS_TRIGGERED)
```
JoseCalero, line 408: the same here

sw/device/lib/drivers/gpio/gpio.c
```diff
@@ -1,213 +1,525 @@
...
+gpio_result_t gpio_intr_check_stat_lvl_high (gpio_pin_number_t pin, bool *is_pending)
+{
+    if (pin > (MAX_PIN-1) || pin < 0)
+    {   
+        *is_pending = GPIO_INTR_IS_NOT_TRIGGERED;
+        return GpioPinNotAcceptable;
+    }
+        
+    if (bitfield_read(gpio_peri->INTRPT_LVL_HIGH_STATUS0, MASK_1b, pin) == 
+        GPIO_INTR_IS_TRIGGERED)
```
JoseCalero, line 440: the  same here

sw/device/target/pynq-z2/x-heep.h
```diff
@@ -16,6 +16,12 @@ extern "C" {
 #define UART_BAUDRATE 115200
 #define TARGET_PYNQ_Z2
 
+/**
+ * As the hw is configurable, we can have setups with different number of
+ * Gpio pins
+ */ 
+#define MAX_PIN     32
```
JoseCalero, line 23: you also need to add this into the same file under sim folder, otherwise, the simulation flow does not compile

sw/device/lib/drivers/gpio/gpio.h
JoseCalero: please, use uint8_t

sw/device/lib/drivers/gpio/gpio.h
```diff
@@ -1,282 +1,377 @@
...
+
+/**
+ * @brief reading from a gpio pin, which is done by reading from GPIO_IN reg
+ * @param gpio_pin_number_t specify pin number
+ * @return gpio value as GpioLow (false) or GpioHigh (true)
+ */
+gpio_result_t gpio_read (gpio_pin_number_t pin, bool *val);
+
+// /**
+//  * @brief set a pin as high. Using masking through GPIO_SET, and then writing
```
JoseCalero, line 193: if this is not needed, please, delete it

sw/device/lib/drivers/gpio/gpio.h
JoseCalero: are they?

JoseCalero (changes requested):
Please, fix the unresolved comments

sw/device/lib/drivers/gpio/gpio.c
Ho3eintaji: I used but it was not working properly so I decided to do it by checking and toggling. However, probably it was my fault, if its better to be implemented in that way, I will check it.

sw/device/lib/drivers/gpio/gpio.c
Ho3eintaji: I added it in x-heep.h, however, im not checking at the moment for more than 32b. Should I extend all the functionalities for the future with checking the define in the functions?!

sw/device/lib/drivers/gpio/gpio.c
JoseCalero: ok

sw/device/lib/drivers/gpio/gpio.c
JoseCalero: let's solve this question tomorrow

Ho3eintaji:
@JoseCalero should be completed now

JoseCalero:
can you do a quick check about the conflicts on "example_power_gating_core/main.c" (it is due to one of the latest merges in main)

Ho3eintaji:
> can you do a quick check about the conflicts on "example_power_gating_core/main.c" (it is due to one of the latest merges in main)

done

JoseCalero:
@Ho3eintaji i have done a check for all the apps, both in compilation and run mode. 
For the freertos blinky app, I think that something is happening with the GPIOs, as only the red one is activated, can you check it?

JoseCalero (changes requested):
can you run freertos app?

JoseCalero:
> can you run freertos app?

ok, i know what is happening, we modified the xdc constraint file for the pynq when we did the PDM integration, so we used pins 21 and 22. So, @Ho3eintaji no need to test it, i will modify the constraint file and the freertos app

davideschiavone:
> > can you run freertos app?
> 
> ok, i know what is happening, we modified the xdc constraint file for the pynq when we did the PDM integration, so we used pins 21 and 22. So, @Ho3eintaji no need to test it, i will modify the constraint file and the freertos app

hi @JoseCalero - if you are using GPIOs 20 and 21, you do not need to modify the constraints file, just set the multiplexer to "GPIO mode" rather than "I2S mode".

```
    pad_control_t pad_control;
    pad_control.base_addr = mmio_region_from_addr((uintptr_t)PAD_CONTROL_START_ADDRESS);
    pad_control_set_mux(&pad_control, (ptrdiff_t)(PAD_CONTROL_PAD_MUX_I2S_SCK_REG_OFFSET), 1);
    pad_control_set_mux(&pad_control, (ptrdiff_t)(PAD_CONTROL_PAD_MUX_I2S_WS_REG_OFFSET), 1);
```


In practice, in `sw/./device/lib/drivers/pad_control/pad_control_regs.h`, the `make mcu-gen` generates the PAD configurations, which is defined [here](https://github.com/esl-epfl/x-heep/blob/main/pad_cfg.hjson).

As you can see [here](https://github.com/esl-epfl/x-heep/blob/main/pad_cfg.hjson#L121), the `GPIO 20` goes now to the PAD `i2s_sck` multiplexed with the signal `i2s_sck` coming from the I2S.
By default, the PAD gets the I2S signal, so you need to set the multiplexer to "GPIO mode" (as above) to tell the PAD to get the GPIO signal instead (in the example above we set both the PAD I2S_SCK and I2S_WS to GPIO mode to get both gpio 20 and 21).

JoseCalero:
@Ho3eintaji can you add those changes to the freertos app?

Ho3eintaji:
> @Ho3eintaji can you add those changes to the freertos app?

Yeah, I will change it.

sw/applications/blinky_freertos/main.c
JuanSapriza, line 1: When the I2S was introduced, it decided to use GPIOs 20-22 (because they were the last ones). 
These were already mapped to the RGB LED5, so they were re-mapped, so GPIOs 15-17 were assigned instead (in the `pin_assign.xdc`. 
This change was never reflected in the `example_freertos_blinky`, which continued to use GPIOs 20-22. 

The only change needed for it to work is to change the `#define GPIO_LD5_*  xx` to their proper numbers.

I will make a PR to this branch with this change and a merge of main.

JuanSapriza:
Just merged with main and fixed conflicts. Tested involved apps in verilator and are working. 

I did a pass over the HAL looks good to me. 
@davideschiavone can we merge this?

Ho3eintaji:
> Just merged with main and fixed conflicts. Tested involved apps in verilator and are working.
> 
> I did a pass over the HAL looks good to me. @davideschiavone can we merge this?

Great, thanks @JuanSapriza 
let's see if it is finally gonna merge or not 👀

PR #247, opened 2023-04-07
fix parameters when there are no external domains

davideschiavone:


davideschiavone:
let me know if this works @Sixardyh

Sixardyh:
Changes tested with both VCS and QuestaSim, works as expected with both simulators

PR #256, opened 2023-04-24
add fast interrupt enable register

fimtrey:
Adding the possibility to disable fast interrupts.

**Minimal prove of concept**
1. Running the adapted dma_example in Verilator works. As soon as we disable the fast_interrupt there is no interrupt triggered (`fast_intr_o = 0`) when an interrupt from the dma arrives (`fast_intr_i[3] = 1'b1`)
image

2. Same on the FPGA, the Interrupt isn't triggered when disabled.


davideschiavone (changes requested):

fimtrey:
> why did you close it? @fimtrey
Tried to clean up my local branches and deleted this branch by accident. Took me a while to extract those commit from the other branch.

PR #259, opened 2023-05-11
External peripheral app bug fix and refactor

StefanoAlbini96:
- Removed useless printf from example_external_peripheral app
- Fixed wrong naming of PLIC function calls in example_external_peripheral app
- Fixed a bug in PLIC for enabling IRQs with index greater then 31
- Add external interrupts handler and macros in the PLIC
- Fixed the behaviour of example_external_peripheral app


JoseCalero:
let's not merge this till the bug for the app is found

JoseCalero (approved):

PR #264, opened 2023-05-15
crt logic update + CMake backend modifications

JoseCalero:
- Based on preprocessor vars. now we have only ONE crt0
- we still need at least two vector files
- minor updates to the CMake back-end to make everything run


JoseCalero:
Yes, the one for the GPIO should be fixed in the GPIO PR. And the one for the external peripheral is in the hospital visiting the bug-doctor

PR #270, opened 2023-05-22
integration of an AMS peripheral example

Sixardyh:
Hello @davideschiavone, here is the PR I was referring to earlier by Slack. No full AMS documentation or .sp files with sky130 yet, but

- a few changes to some markdown files (mostly mentioning Verdi can be used for waveforms, and also a small restructuration of the IntegratePeripheral.md file as it had a few incoherences IMO)
- the various ams peripheral files and its integration into heep
- a dummy .sv file for the adc for when digital-only simulations are made
- a modification of Makefile/the main .core to separate digital from AMS simulation. (make vcs-ams-sim versus make vcs-sim)

Both vcs-sim & vcs-ams-sim / digital & mixed-signal have been tested.


hw/core-v-mini-mcu/ao_peripheral_subsystem.sv
```diff
@@ -410,5 +410,14 @@ module ao_peripheral_subsystem
       .intr_rx_parity_err_o(uart_intr_rx_parity_err_o)
   );
 
+  ams #(
```
davideschiavone, line 413: we should be better plug it into the testharness

davideschiavone (changes requested):

davideschiavone (changes requested):
left commens on code

davideschiavone (changes requested):
left commens on code

core-v-mini-mcu.core
```diff
@@ -32,6 +32,7 @@ filesets:
     - x-heep:ip:power_manager
     - x-heep:ip:fast_intr_ctrl
     - x-heep:ip:pdm2pcm
+    - x-heep:ip:ams
```
davideschiavone, line 35: this should be part of the ip_example

davideschiavone (changes requested):

x-heep-tb-utils.core
```diff
@@ -1,6 +1,6 @@
 CAPI=2:
 
-# Copyright 2022 OpenHW Group
+ mpleCopyright 2022 OpenHW Group
```
davideschiavone, line 3: can you pls fix this?

hw/analog/adc.sp
```diff
@@ -0,0 +1,94 @@
+** Copyright EPFL contributors.
```
davideschiavone, line 1: please move the analog folder under ip_examples/ams , so it should be

`hw/ip_examples/ams/analog`

davideschiavone (changes requested):

Issue #276, opened 2023-05-24
Power gating core app needs to use different GPIOs


JuanSapriza:
Currently the `example_power_gating_core` application uses GPIOs 30 and 31 to trigger an interrupt to wake the core up. 

Those GPIOs are no longer accessible in the FPGA (I think they were used for the SPI2). The app should be modified to consider this new situation.

davideschiavone:
this part of the application should just be removed when `#ifndef PYNQ` as it is just for the testbench

@simone-machetti pls check that works on Verilator and add the #ifndef to the latest part pls :)

davideschiavone:
this is fixed now

PR #286, opened 2023-05-25
fix example_power_gating_core app

simone-machetti:
Solves issue #276. 

The app had some bugs that have been fixed and now works well in simulation (tested with Verilator). However, it is still not working on the Pynq-Z2 board (the app never ends).


davideschiavone:
must work on FPGA before merging

simone-machetti:
Mergeable.

davideschiavone:
> Mergeable.

does it work on the FPGA ?

simone-machetti:
Yes, I tested it.

PR #288, opened 2023-05-26
add ifdef peripheral included

davideschiavone:


sw/applications/example_external_peripheral/main.c
```diff
@@ -25,6 +25,10 @@ plic_irq_id_t intr_num;
 int main(int argc, char *argv[])
 {
 
+#ifndef RV_PLIC_IS_INCLUDED
+  #pragma message ( "This app does NOT work as the RV_PLIC peripheral is not included" )
```
JoseCalero, line 29: I would output an error, so the app is not even compiled. If we do not have the PLIC for example, why do we want the app to be able to be compiled? so:
`#pragma error ( "This app does NOT work as the RV_PLIC peripheral is not included" )`
or
`#error ( "This app does NOT work as the RV_PLIC peripheral is not included" ) `
Both work

JoseCalero (changes requested):
let's avoid even the possibility to compile those

sw/applications/example_external_peripheral/main.c
JuanSapriza: We have to be careful with this or the CI might reject every push. We might need to compile for both targets and make a logic OR maybe?

davideschiavone: yes I kept it like this as otherwise the application does not end, maybe it is better to return 0 rather than -1 as indeed the CI may complain that application which is not supposed to run fail

JuanSapriza: Having talked with Jose we agree that an error would be a good solution.

davideschiavone: and what about the CI?

JoseCalero: so, we want the application to be able to run even though we do not have (for example) the PLIC included?

davideschiavone: no but the CI compiles anyway all the program and a future CI task may run all the applications, so we need to find a way that the CI does not report "errors" -

JuanSapriza: But if I push an application that requires a change in the `mcu_cfg.hjson` to compile, I think the CI should raise an error.

JoseCalero: unless i am missing something, I agree

davideschiavone: done @JuanSapriza @JoseCalero

davideschiavone:
the CI fails as discussed above, the PDM2PCM peripheral is not included, hence the error when trying to compile it - either we do not compile them all, or another solution is needed

JuanSapriza:
> the CI fails as discussed above, the PDM2PCM peripheral is not included, hence the error when trying to compile it - either we do not compile them all, or another solution is needed

I guess that the pcd2pcm is not included because a) it is not ready yet, or b) It a feature specific to a particular domain. 

If the case is a), I understand it should have its own branch, and not be pushed to main. 
If the case is b), maybe we should include it and add a step in the configuration of the project suggesting to go and disable whatever is not needed.

davideschiavone:
> > the CI fails as discussed above, the PDM2PCM peripheral is not included, hence the error when trying to compile it - either we do not compile them all, or another solution is needed
> 
> I guess that the pcd2pcm is not included because a) it is not ready yet, or b) It a feature specific to a particular domain.
> 
> If the case is a), I understand it should have its own branch, and not be pushed to main. If the case is b), maybe we should include it and add a step in the configuration of the project suggesting to go and disable whatever is not needed.

x-heep is configurable, so it allows to have or not a given peripheral by modifying the mcu-gen - so by default, the pdm2pcm is not included (for whatever reason) so being in the main branch is a must, so we need to find another solution

JoseCalero:
> > > the CI fails as discussed above, the PDM2PCM peripheral is not included, hence the error when trying to compile it - either we do not compile them all, or another solution is needed
> > 
> > 
> > I guess that the pcd2pcm is not included because a) it is not ready yet, or b) It a feature specific to a particular domain.
> > If the case is a), I understand it should have its own branch, and not be pushed to main. If the case is b), maybe we should include it and add a step in the configuration of the project suggesting to go and disable whatever is not needed.
> 
> x-heep is configurable, so it allows to have or not a given peripheral by modifying the mcu-gen - so by default, the pdm2pcm is not included (for whatever reason) so being in the main branch is a must, so we need to find another solution

Given that the PDM2PCM peripheral still needs some work I would come back then for the warnings implementations, just as a temporary solution (although the warning is gonna be 'lost' in the looong output of the CI). To avoid this type of problem in the future, there are different options/possibilities:

1. avoid merging peripherals/apps (whatever) that do not work/haven't been tested properly. (As the project gets bigger, this is gonna be challenging)
2. X-HEEP is configurable (e.g.: yes/no included peripherals), but if by default (main branch) we do not include a peripheral, maybe its application should be placed in another folder? (CI is gonna compile just the ones that are under the applications folder).
3. forget about 1 and 2, and just add warnings.

what do you think?

davideschiavone:
> > > > the CI fails as discussed above, the PDM2PCM peripheral is not included, hence the error when trying to compile it - either we do not compile them all, or another solution is needed
> > > 
> > > 
> > > I guess that the pcd2pcm is not included because a) it is not ready yet, or b) It a feature specific to a particular domain.
> > > If the case is a), I understand it should have its own branch, and not be pushed to main. If the case is b), maybe we should include it and add a step in the configuration of the project suggesting to go and disable whatever is not needed.
> > 
> > 
> > x-heep is configurable, so it allows to have or not a given peripheral by modifying the mcu-gen - so by default, the pdm2pcm is not included (for whatever reason) so being in the main branch is a must, so we need to find another solution
> 
> Given that the PDM2PCM peripheral still needs some work I would come back then for the warnings implementations, just as a temporary solution (although the warning is gonna be 'lost' in the looong output of the CI). To avoid this type of problem in the future, there are different options/possibilities:
> 
> 1. avoid merging peripherals/apps (whatever) that do not work/haven't been tested properly. (As the project gets bigger, this is gonna be challenging)
> 2. X-HEEP is configurable (e.g.: yes/no included peripherals), but if by default (main branch) we do not include a peripheral, maybe its application should be placed in another folder? (CI is gonna compile just the ones that are under the applications folder).
> 3. forget about 1 and 2, and just add warnings.
> 
> what do you think?

1. The test for the pdm2pcm peripheral works, it is just that the peripheral it is not included by default - so it can even be that another peripheral (say spi2) gets excluded for area reason, then we could come back to this problem anyway for apps that use the spi2

2. I do not like tbh

3. the error is ok - I think we should just have a CI_list.txt instead of compiling all the possible applications - this CI_list.txt should be carefully updated and maintained of course, but at least we check only what we want

JuanSapriza:
>     3. the error is ok - I think we should just have a CI_list.txt instead of compiling all the possible applications - this CI_list.txt should be carefully updated and maintained of course, but at least we check only what we want

What I dont like about this is that we will not be testing this app in the CI. If some change brakes it, we might not find out.

davideschiavone:
> > (Thinking out loud). One thing we could have is some sort of configuration/requirements-file per app. Before compiling we could check that file, compare to what is already generated and, if there is a difference, make mcu-gen with the proper configuration.
> 
> @davideschiavone Do you think this would be too much work? I think it would be the better option in the mid-term

It is easier if we enable all the peripherals by default and let others enable/disable them passing a different mcu.hjson file - which could also make sense to me (as we already did with the PADs) - shall we do this way?

davideschiavone:
> Could we add a parameter in the mcu-gen that determines which mcu-cfg to use? (`mcu_cfg_full.hjson` or `mcu_cfg_light.hjson`). By default we can grab the light one, but in the CI we can override this and always compile with the full version. This way, for the average user its transparent.
> 
> Furthermore, the full version could use the light one as base and just add some more components (as a form of patch)? (I might be delusional here)

actually is a good idea, but of I may, I would do an mcu_cfg_ci and and mcu_cfg. What do you think ?

JuanSapriza:
Indeed that's the easiest solution! 
The CI could be updated as:
```bash
sed 's/is_included: "no",/is_included: "yes",/' mcu_cfg.hjson
make mcu-gen
```
I will test in this branch. If everything goes fine, I can make a PR containing that change.

JoseCalero:
> Indeed that's the easiest solution! The CI could be updated as:
> 
> ```shell
> sed 's/is_included: "no",/is_included: "yes",/' -i mcu_cfg.hjson
> make mcu-gen
> ```
> 
> I tested this change in this PR by reverting the changes of the `CI_list.txt` and it works.
> 
> I made PR #293. Once merged this PR needs to revert those changes as well and then can be merged.

#293 was merged.

davideschiavone:
> > Indeed that's the easiest solution! The CI could be updated as:
> > ```shell
> > sed 's/is_included: "no",/is_included: "yes",/' -i mcu_cfg.hjson
> > make mcu-gen
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I tested this change in this PR by reverting the changes of the `CI_list.txt` and it works.
> > I made PR #293. Once merged this PR needs to revert those changes as well and then can be merged.
> 
> #293 was merged.

Merged back in this branch - I would say it is ready to be merged @JoseCalero @JuanSapriza

Issue #313, opened 2023-06-21
FuseSoC parameters without flags


StMiky:
## Description

_Requested by @davideschiavone_

_X-HEEP_'s main `.core` file use FuseSoC flags to set build parameters (Verilog parameters) instead of directly including them in the target. This extra step is unnecessary since FuseSoC supports parameter definiton directly in the command-line (as shown by `fusesoc run ... openhwgroup.org:systems:core-v-mini-mcu --help`). Avoiding flag definition simplifies the command line to launch FuseSoC target rules.

PR #314, opened 2023-06-21
Directly use FuseSoC parameters (no flags)

StMiky:
## Description
Tentative solution to #313.

## Status
**Not** properly tested yet and not applicable for Verilog `define`s or flags that are also used to conditionally include filesets.


StMiky:
Issue handled by @davideschiavone from now on.

Issue #311, opened 2023-06-16
[hal] Modification in interrupt handler assignment


JuanSapriza:
Currently, the interrupt handlers are defined as weak functions in `plic.c`:
```C
__attribute__((weak, optimize("O0"))) void handler_irq_uart( void ) { } 
```
Added to an array of function pointer:
```C
handler_funct_t handlers[PLIC_INTR_SRCS_NUM] = {&handler_irq_uart, 
                                                &handler_irq_gpio, 
                                                &handler_irq_i2c, 
                                                &handler_irq_spi,
                                                &handler_irq_ext};
```
and then, when an interrupt happens, the PLIC checks which pin triggered the interrupt. Then locates that pin inside one of the possible ranges: 
```C
...
#define I2C_ID_START    33
#define I2C_ID_END      48
#define SPI_ID          49
#define EXT_IRQ_START   50
#define EXT_IRQ_END     63
```

Each of which correspond to one of the handler functions. The function is then called through the pointer 
```C
void handler_irq_external(void){
  plic_irq_id_t int_id = 0;
  plic_result_t res = plic_irq_claim(&int_id);
  irq_sources_t type = plic_get_irq_src_type(int_id);
  if(type != IRQ_BAD) {
    handlers[type]();               // HERE THE FUNCTION IS CALLED
    plic_intr_flag = 1;
    plic_irq_complete(&int_id);
  }
}
```

The weak implementations can then be redefined in the application to attend the interrupt.

A problem appears when the application uses an external peripheral. The `handler_irq_ext` has no way of determining which pin was the source of the interrupt and, therefore, if it corresponds to it. 

One possible solution is to pass the id of the pin to the handler: 
```C
void handler_irq_external(void){
  plic_irq_id_t int_id = 0;
  plic_result_t res = plic_irq_claim(&int_id);
  irq_sources_t type = plic_get_irq_src_type(int_id);
  if(type != IRQ_BAD) {
    handlers[type]( int_id );       // THE INTERRUPT ID IS PASSED
    plic_intr_flag = 1;

    plic_irq_complete(&int_id);
  }
}
```

A minor inconvenience of this solution is that it requires every handler to take a parameter (to use the function-pointer-array), despite they will not use it. 
```C
__attribute__((weak, optimize("O0"))) void handler_irq_uart(plic_irq_id_t id) { } 
```

[...]

PR #315, opened 2023-06-22
Interrupt ID to the intr handlers of the PLIC

JuanSapriza:
Because some interrupt handlers need to know what the source of the interrupt was, all handlers needed to take a parameter with the interrupt ID. 
That was added and changed were necessary. 

This was of interest to @StMiky 
@StefanoAlbini96 could you review?

### DIFF
--- a/sw/applications/example_gpio_cnt/main.c
+++ b/sw/applications/example_gpio_cnt/main.c
@@ -49,7 +49,7 @@ Notes:
 // plic_params_t rv_plic_params;
 // plic_t rv_plic;
 plic_result_t plic_res;
-// plic_irq_id_t intr_num;
+// uint32_t intr_num;
 
 // void handler_irq_external(void) {
 //     // Claim/clear interrupt
--- a/sw/applications/example_power_gating_core/main.c
+++ b/sw/applications/example_power_gating_core/main.c
@@ -157,7 +157,7 @@ int main(int argc, char *argv[])
     }
     CSR_SET_BITS(CSR_REG_MSTATUS, 0x8);
 
-    plic_irq_id_t intr_num;
+    uint32_t intr_num;
     plic_irq_complete(&intr_num);
 #endif
 
--- a/sw/applications/example_virtual_flash/main.c
+++ b/sw/applications/example_virtual_flash/main.c
@@ -24,7 +24,7 @@
 
 // Interrupt controller variables
 plic_result_t plic_res;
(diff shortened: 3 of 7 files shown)


JuanSapriza:
This cannot be built with clang :upside_down_face: 

```
 In file included from /__w/x-heep/x-heep/sw/applications/spi_flash_write/main.c:13:
/__w/x-heep/x-heep/sw/device/lib/drivers/rv_plic/rv_plic.h:103:33: error: a parameter list without types is only allowed in a function definition
typedef void (*handler_funct_t)(plic_irq_id_t);
```

It seems that CLANG messes up with the `typedef` of the `plic_irq_id_t` as the parameter. To overcome this, I removed the `typedef uint32_t plic_irq_id_t` and used plainly `uint32_t`. This builds in both cases. 

@StefanoAlbini96 please confirm if this is still acceptable.

JuanSapriza:
Related to #311

StMiky:
Thank you @JuanSapriza. I tested this on my system with external interrupts and had no issues :v:

StMiky:
> Thank you @JuanSapriza. I tested this in my system with external interrupts and had no issues v

Actually I had to patch the number of sources (`PLIC_INTR_SRCS_NUM`) from 5 to 6 in [`rv_plic.h`](https://github.com/JuanSapriza/x-heep/blob/6b16e84b69c7bd2fd19a716186bcd288bbd6a139/sw/device/lib/drivers/rv_plic/rv_plic.h#L56) for the compiler to actually map `handler_irq_ext` in `handlers`.

JuanSapriza:
> Actually I had to fix the number of sources (`PLIC_INTR_SRCS_NUM`) from 5 to 6 in [`rv_plic.h`](https://github.com/JuanSapriza/x-heep/blob/6b16e84b69c7bd2fd19a716186bcd288bbd6a139/sw/device/lib/drivers/rv_plic/rv_plic.h#L56). To avoid similar issues in the future, and unless that macro is necessary somewhere else, a possibility is to omit it as it is not strictly necessary for the definition of the `handlers[]` array.

Indeed, this how I had it, but forgot to make this change on the PR. Good catch! I will update it ASAP

JuanSapriza:
Will update changes from main and re-submit

JuanSapriza:
@StefanoAlbini96 this is ready for you review

StefanoAlbini96 (approved):

PR #320, opened 2023-06-23
add ports to DMA to write to addresses coming from port2 of DMA

davideschiavone:


davideschiavone:
@JuanSapriza , @JoseCalero - the APP does not compile if MEMORY_BANKS < 3 - what can we do?

sw/device/lib/drivers/dma/dma.c
```diff
@@ -697,11 +706,14 @@ dma_config_flags_t dma_load_transaction( dma_trans_t *p_trans )
...
-                    DMA_PTR_INC_DST_PTR_INC_MASK,
-                    DMA_PTR_INC_DST_PTR_INC_OFFSET );
+
+    if(dma_cb.trans->mode != DMA_TRANS_MODE_ADDRESS) {
+        //printf("dst inc - ");
+        write_register(  get_increment_b( dma_cb.trans->dst ),
+                        DMA_PTR_INC_REG_OFFSET,
+                        DMA_PTR_INC_DST_PTR_INC_MASK,
+                        DMA_PTR_INC_DST_PTR_INC_OFFSET );
+    }
```
JuanSapriza, line 716: I understand that in _ADDRESS MODE_ the destination increment is ignored right? 
I will add this to the documentation, but can we add a comment here to make it explicit?

sw/device/lib/drivers/dma/dma.c
```diff
@@ -675,8 +675,17 @@ dma_config_flags_t dma_load_transaction( dma_trans_t *p_trans )
...
-    dma_cb.peri->DST_PTR = dma_cb.trans->dst->ptr;
-    
+
+    if(dma_cb.trans->mode != DMA_TRANS_MODE_ADDRESS) {
+        /*
+            Write to the destination pointers only if we are not in address mode,
+            otherwise the destination address is read in a separate port in parallel with the data
+            from the address port
+        */
+        dma_cb.peri->DST_PTR = dma_cb.trans->dst->ptr;
```
JuanSapriza, line 685: The agreed format in the FW team was to use the curly braces in their own line. We should correct these occurrences for consistency (at least in the HAL, the examples are not so consistent on this, they should go on a refactor at some point)

JuanSapriza (changes requested):
Just leaving two minor comments from the format point fo view

sw/device/lib/drivers/dma/dma.c
davideschiavone: correct

sw/device/lib/drivers/dma/dma.c
davideschiavone: how shall I modify them?

JuanSapriza:
This was tested on both FPGA (flash_load) and Verilator (even without the PRINTFs it takes a while) and works just fine. 
It should be good for merging straight away after the requested changes.

sw/device/lib/drivers/dma/dma.c
JuanSapriza: We should do: 
```C
    if(dma_cb.trans->mode != DMA_TRANS_MODE_ADDRESS) 
    {
        /*
            Write to the destination pointers only if we are not in address mode,
            otherwise the destination address is read in a separate port in parallel with the data
            from the address port
        */
        dma_cb.peri->DST_PTR = dma_cb.trans->dst->ptr;
     } 
     else 
     {
        dma_cb.peri->ADDR_PTR = dma_cb.trans->src_addr->ptr;
     }
```

PR #322, opened 2023-06-28
refactoring of examples

JuanSapriza:
During the testing of all example apps we faced several minor issues that can be tackled without affecting each other. We will be assessing those issues in this PR. 
So far this includes: 

* Changing occurrences of `printf` for `PRINTF` so it can be disabled by default for simulation and enabled for FPGA: 
```C
/* Change this value to 0 to disable prints for FPGA and enable them for simulation. */
#define DEFAULT_PRINTF_BEHAVIOR 1

#ifdef TARGET_PYNQ_Z2 
    #define ENABLE_PRINTF DEFAULT_PRINTF_BEHAVIOR
#else 
    #define ENABLE_PRINTF !DEFAULT_PRINTF_BEHAVIOR
#endif

#if ENABLE_PRINTF
#else
  #define PRINTF(...)
#endif 
```

* Add `\r` on PRINTFs to better visualize them on `screen`.
* Rename example apps so they follow the standard `example_*`.
* Reduced the size of large buffers in `dma` and `i2s` examples so they don't require more memory banks. 
* Standardized some PRINTF messages of `success` and `failure`


Makefile
```diff
@@ -51,6 +51,10 @@ ARCH     ?= rv32imc
 # Path relative from the location of sw/Makefile from which to fetch source files. The directory of that file is the default value.
 SOURCE 	 ?= "."
 
+# Number of memory banks. Set to 3 by default to allow the testing of ram retention examples
+MEMORY_BANKS 	?= 3
```
davideschiavone, line 55: this I cannot accept it :) must be 2 otherwise we break compatibility with existing systems

davideschiavone (changes requested):

Makefile
JuanSapriza: You mean any project should now modify their external Makefile to force mcu-gen to include `MEMORY_BANKS=2`. 
I can agree thats far from nice. 

Options I see are:  
1. We use 3 mb by default (requires project depending on the 2 mb to make a change on their external Makefile) 
2. We find a way of making the `set_retentive` apps work with 2 mb (no idea if this is even feasible) 
3. We make all tests with `MEMORY_BANKS=3` (but if someone trying an app as-is -with 2 mb-, it might not work)
4. We test only these apps with `MEMORY_BANKS=3` (it would slow down testing and pose a slight challenge if we try to automate tests)

Opinions?   @davideschiavone @simone-machetti

JuanSapriza: This change was overturned

sw/applications/example_ams_peripheral/main.c
```diff
@@ -19,29 +19,22 @@ int main(int argc, char *argv[])
         unsigned int AMS_START_ADDRESS = EXT_PERIPHERAL_START_ADDRESS + 0x1000;
         mmio_region_t ams_base_addr = mmio_region_from_addr((uintptr_t)AMS_START_ADDRESS);
 
-        printf("0");
```
davideschiavone, line 22: pls keep these printf in this specific example

davideschiavone (changes requested):

sw/applications/example_ams_peripheral/main.c
davideschiavone: compulsory (it is an analog example, we need those...)

JuanSapriza: done

sw/applications/example_set_retentive_external_ram_blocks/main.c
```diff
@@ -1,5 +1,5 @@
 // Copyright EPFL contributors.
-// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.\n
```
davideschiavone, line 2: shouldn't we change this application to return EXIT_SUCCESS in case NUM_BANKS = 2? 

AFAIK, the CI uses MEMBANKS = 3, can we change it to 2?

sw/applications/example_dma/main.c
```diff
@@ -18,25 +18,31 @@
 #define TEST_ADDRESS_MODE_EXTERNAL_DEVICE
 
 #define TEST_DATA_SIZE      16
-#define TEST_DATA_LARGE     4096
+#define TEST_DATA_LARGE     1024
```
davideschiavone, line 21: you can put back 4096 if you want, or does it compile with 2 banks like this?

in case it does not, please guard the application to run only with MEMBANKS > 2

otherwise warning and report 0

davideschiavone: nice

davideschiavone:
let me know when I need to review it, do not set it as a `draft PR` when it is ready @JuanSapriza

sw/applications/example_set_retentive_external_ram_blocks/main.c
JuanSapriza: The CI is currently using 6 (at the time freertos required 5). It can be reduced to 2. We could make an issue, but we might want to review this before going for the simulation CI. 

Regarding this app, this works with 2mb (my comment at the top of this PR was wrong). 

Also, good catch with that `\n`

JuanSapriza:
> let me know when I need to review it, do not set it as a `draft PR` when it is ready @JuanSapriza

It should be good now. I was planning on leaving this as a draft until we finish all tests, but we can merge and open a new one if we need.

davideschiavone (commented):
we need to change the set_retentive app

sw/applications/example_set_retentive_external_ram_blocks/main.c
JuanSapriza: I tried it with 2mb and it worked properly and returned `SUCCESS`

davideschiavone: but here it is written return failire https://github.com/JuanSapriza/x-heep/blob/bf52f256048d5ad0e904f58debef9d86673e9407/sw/applications/example_set_retentive_ram_blocks/main.c#L70

JuanSapriza: but this is the example set retentive EXTERNAL ram blocks, which i guess makes sense it does not have this issue. 

What i can do is change the name to `example_set_retentive_ram_blocks_external` so the difference is left at the end of the name to avoid further confusions

JuanSapriza: Done

PR #325, opened 2023-06-30
fix several applications

davideschiavone:


sw/applications/example_matfadd/main.c
```diff
@@ -64,6 +144,18 @@ void __attribute__ ((noinline)) matrixAdd(float *  A, float *  B, float *  C, in
     }
 }
 
+void __attribute__ ((noinline)) printMatrix(float *  C, int N, int M)
```
JuanSapriza, line 147: Because this definition is before the function call, it gives an error in clang. If you move the definition or make a declaration above main, it should be good

JuanSapriza (approved):
All good for me

PR #339, opened 2023-07-05
fix #338

### LINKED ISSUE #338: bug in applications to wait for interrupts
We wrote

```
    while ( ! CONDITION ) {
        // disable_interrupts
        // this does not prevent waking up the core as this is controlled by the MIP register
        CSR_SET_BITS(CSR_REG_MSTATUS, 0x0);
        if ( CONDITION == 0 ) {
            wait_for_interrupt();
            //from here we wake up even if we did not jump to the ISR
        }
        CSR_SET_BITS(CSR_REG_MSTATUS, 0x8);
    }
```

should be

```
    while ( ! CONDITION ) {
        // disable_interrupts
        // this does not prevent waking up the core as this is controlled by the MIP register
        CSR_CLEAR_BITS(CSR_REG_MSTATUS, 0x8);
        if ( CONDITION == 0 ) {
            wait_for_interrupt();
            //from here we wake up even if we did not jump to the ISR
        }
        CSR_SET_BITS(CSR_REG_MSTATUS, 0x8);
    }
```

`CSR_CLEAR_BITS(CSR_REG_MSTATUS, 0x8);` should be used to disable interruipts

davideschiavone:


davideschiavone:
@JuanSapriza -  I tried the 3 applications on Verilator and they work as expected

JuanSapriza (approved):
I tested on FPGA `example_dma` and `example_gpio_intr` for both `flash_load` and `flash_exec`. Both work correctly. 

Later we should change the 10 cases where a `wait_for_interrupt()` is called without the interrupt disable/enable safeguard.

PR #349, opened 2023-07-18
adding attributes and mux pad parameters in mcu-gen

davideschiavone:
With and without pad attributes:

- [x] try questasim
- [x] try vcs
- [x] try verilator (run all apps)
- [x] try heeperator
- [x] try vivado


JuanSapriza:
All apps where tested with verilator using the `test_all.sh` script from #341, only the SPI ones did not work (as expected).

PR #351, opened 2023-07-20
moved fpnew in its own directory

davideschiavone:


hw/vendor/pulp_platform_fpnew.core
```diff
@@ -0,0 +1,48 @@
...
+
+# Copyright 2021 OpenHW Group
+# Solderpad Hardware License, Version 2.1, see LICENSE.md for details.
+# SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+
+name: "pulp-platform.org::fpnew"
+description: "PULP Platform RISC-V FPU"
+
+filesets:
+  files_rtl:
```
davidmallasen, line 11: I think this should be added here:
~~~
    depend:
    - pulp-platform.org::common_cells
~~~

davidmallasen (approved):

Issue #347, opened 2023-07-18
Change active low signal names in power manager with `*_ni` or `*_no` suffixes


StMiky:
## Example
`set_retentive_o` is active low, so it should be named `set_retentive_no`. 

@simone-machetti

davideschiavone:
294a968f7934425d2eb251527e3a217a2f38e5d2

PR #352, opened 2023-07-20
change signal names in power manager to reflex polarity

simone-machetti:


simone-machetti:
Fixes issue #347.

davideschiavone:
I let @StMiky review it, but I guess this should be done for each signal active low (if it is not the only one)

simone-machetti:
Yes, I think it is better. I'll fix it.

davideschiavone:
we should check that works with the UPF simulation,  so please make sure you modify the `upf` as well

simone-machetti:
True.

simone-machetti:
Tested UPF with Questasim.

davideschiavone:
this is not working, for example, the application `example_cpu_powergate` is not working, a quick look shows that the isolation signal for example has never been asserted. Did you try this application @simone-machetti ?

davideschiavone:
the x-heep wrapper for vivado is  also broken

davideschiavone:
> this is not working, for example, the application `example_cpu_powergate` is not working, a quick look shows that the isolation signal for example has never been asserted. Did you try this application @simone-machetti ?

it actually does not work on the `main` branch either, so something has changed

simone-machetti:
Fixed.

Issue #336, opened 2023-07-04
Include extenal C runtime files in the CMake flow


danivz:
The current CMake software compilation flow doesn't support external C runtime files.

Following the instructions in [eXtending_HEEP.md](https://github.com/esl-epfl/x-heep/blob/8f2351053dd240d4e6a01ac90cfcd49fb78a747c/eXtendingHEEP.md) I have:
```
BASE
    └── sw
        ├── applications
        │   └── your_app
        │       ├── main.c
        │       ├── your_app.c
        │       ├── your_app.h
        │       └── ...
        ├── build -> ../hw/vendor/esl_epfl_x_heep/sw/build
        ├── device -> ../hw/vendor/esl_epfl_x_heep/sw/device
        ├── linker -> ../hw/vendor/esl_epfl_x_heep/sw/linker
        └── external
            └── lib
                ├── crt
                |   └── external_crt0.S
                └── drivers
                │       └── fll
                |           └── fll_regs.h
                └── runtime
                    └── heepie.h
```

I want to include `external_crt0.S` in https://github.com/esl-epfl/x-heep/blob/8f2351053dd240d4e6a01ac90cfcd49fb78a747c/sw/CMakeLists.txt#L169C1-L172C37. However, If I hardcode the CRT file path in `LINKED_FILES` the compilation fails because `fll_regs.h` and `heepie.h` are included in `external_crt0.S` and the compiler can't find these headers.

I also noticed that in `crt0.S` the macro `EXTERNAL_CRT0` isn't set automatically when there is a external CRT file: https://github.com/esl-epfl/x-heep/blob/8f2351053dd240d4e6a01ac90cfcd49fb78a747c/sw/device/lib/crt/crt0.S#L44C1-L46C7

davideschiavone:
Well catched, this function should be restored asap

JoseCalero:
This should be fixed and closed due to #358

davideschiavone:
@danivz , can you please close this issue and tell us if the solution works?

JuanSapriza:
@JoseCalero you sure #358 would fix this? It did not implement what we discussed in [this comment](https://github.com/esl-epfl/x-heep/issues/336#issuecomment-1652144768)

JoseCalero:
You are right, we are missing the external, I will do it asap

JuanSapriza:
Being fixed in #364

PR #358, opened 2023-07-27
Cmakelists find external apps

JuanSapriza:
Previously, the CMakeLists would look into the list of header files in the `SOURCE_PATH`, if no header with the application's name was found, it would consider the application to be located inside the default X-HEEP `sw/applications` folder. 

This prevented applications without a header file to be built. 

Now there is a distinction between the place where the application is located and the determination that the app was found or not. 

If the application is not found in the provided `SOURCE_PATH` it is marked as _not found_ and a new search is performed inside the default directory by setting `SOURCE_PATH` to `ROOT_PROJECT`. 
  
If by the end of this process `SOURCE_PATH` is different from `ROOT_PROJECT`, then the application was indeed external and the flag is raised (to be used, for example, in the selection of runtime files in #336

### DIFF
--- a/sw/CMakeLists.txt
+++ b/sw/CMakeLists.txt
@@ -54,33 +54,31 @@ SET(MAINFILE "main")
 # Make a list of the header files that need to be included
 FILE(GLOB_RECURSE new_list FOLLOW_SYMLINKS ${SOURCE_PATH}*.h)
 SET(dir_list_str "")
-SET(external_app 0)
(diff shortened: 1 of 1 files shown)


JuanSapriza:
@JoseCalero @davidmallasen, could you test this?

davidmallasen (approved):
Works for me

JoseCalero (approved):

PR #359, opened 2023-07-27
Choose linker compiler and sim/nosim in test_all.sh

### LINKED ISSUE #280: CI to test compilation using differents configurations
Right now, we have a CI that tests all apps using the default configurations (target: sim, linker: on_chip, etc)

We would like to have a script like the one in `X-HEEP/scripts/compile_all_apps.sh`. 

Note that this app has a bug as arrays should be defined as 
```
declare -a arr=("element1" "element2" "element3")
for i in "${arr[@]}"
do
    ....
done
```

I have tested this feature on my local computer using a variation of the script in the CI and it takes just slightly longer than the original one (the huge container configuration overhead is the real bottleneck).

JuanSapriza:
I added the possibility of choosing the linker, compiler and to avoid simulation in the test script. 

Source it like
```bash
. test_all.sh
```
And it will compile only with `gcc`, for `on_chip` and simulate with verilator.

Source it like 
```bash 
. test_all.sh gcc clang flash_load nosim
``` 
And it will compile for both `gcc` and `clang`, for only `flash_load` and perform no simulations. 

---

Although its not a CI, i think it will fix #280


JoseCalero:
I tested this, first simulation was successful, but did not continue with the rest, see:
image

I will check it again tomorrow

JoseCalero (changes requested):
Need to fix option management for different linkers

Issue #305, opened 2023-06-13
Minor modifications in eXtendingHEEP.md


JuanSapriza:
Some changes need to be done in the explanations of `eXtendingHEEP.md`: 

```
MAKE     = make

test:
    @echo Nothing is executed from X-HEEP 
```
Mention that `test` is not a target in X-HEEP  

```
app:
    @echo This is app being run from the x-heep repo
    $(MAKE) -f $(XHEEP_MAKE) $(MAKECMDGOALS) PROJECT=hello_world SOURCE=.
```
Make it clear that `app` WILL BE run from inside the X-HEEP repo.

```
verilator-sim: 
    @echo You will not access verilator-sim
```
Make it clear that the `verilator-sim` that will not be accessed is the one from inside X-HEEP

Furthermore, reflect these changes inside the `external.mk` file, or link to the documentation instead of leaving a comment.

Issue #304, opened 2023-06-13
External makefile rules


davidmallasen:
I have an external makefile as described in `external.mk`. After running a rule that doesn't exist in X-HEEP, but does in the Makefile of the external project, the Makefile of X-HEEP also tries to execute the same rule (and fails because it doesn't exist). If the rule does exist in the externall Makefile, it should stop and not continue running. This could be solved by setting the external rules as `.PHONY` in some cases, but should be investigated further.

JuanSapriza:
This issue can be solved as per #360 @davideschiavone @JoseCalero

PR #360, opened 2023-07-27
Improved comments on the eXtendingHEEP readme

JuanSapriza:
This solves #305.
Note that #304 is not solved as we should look deeper into it. Anyway, I added a comment with the solution we have found so far.


JoseCalero (approved):

davideschiavone:
@JoseCalero - can we merge this?

PR #361, opened 2023-07-27
Simplified PRINTF logic

JuanSapriza:
Previously, the PRINTF logic was as follows;
```C
/* Change this value to 0 to disable prints for FPGA and enable them for simulation. */
#define DEFAULT_PRINTF_BEHAVIOR 1

#ifdef TARGET_PYNQ_Z2 
    #define ENABLE_PRINTF DEFAULT_PRINTF_BEHAVIOR
#else 
    #define ENABLE_PRINTF !DEFAULT_PRINTF_BEHAVIOR
#endif

#if ENABLE_PRINTF
#else
  #define PRINTF(...)
#endif 
```

This was deemed confusing, so it was changed for: 
```C
#define PRINTF_IN_FPGA  1
#define PRINTF_IN_SIM   0

#if TARGET_SIM && PRINTF_IN_SIM
#elif TARGET_PYNQ_Z2 && PRINTF_IN_FPGA
#else
    #define PRINTF(...)
#endif
```

---

This changed was simplified by changing 
```C
#define TARGET_XXX 
```
for
```C
#define TARGET_XXX 1
```
Which has no impact on any `#ifdef`, but allows to perform `#if`. 

@Ho3eintaji could you take a look and see if you agree on the new logic? 

---

This was tested on all apps using verilator, on_chip, gcc, 
On FPGA using flash_load only for `example_matadd` and `hello_world`


JoseCalero:
The implementation is Ok, but this PRINTF logic should be abstracted from the user/app point of view, i.e. this used to be implemented in runtime files (e.g. syscall, common example from Nordic is NRF_LOG_..., Etc). This does not mean we cannot merge it, we can, but it should not be done in the APPs (a matter of abstraction)

PR #265, opened 2023-05-16
i2c HAL refactor

StefanoAlbini96:
Major refactoring of the HAL for the i2c peripheral.
The main changes are due to the usage of the memory mapped `i2c_peri` structure to access directly the registers.
Moreover the code has been modified to be more clear and compact.

PR #363, opened 2023-08-07
PLIC handlers array

JuanSapriza:
# Introduction 

To allow for easier extension of X-HEEP (without the need of modifying the PLIC HAL), we have decided to migrate the pre-defined handlers into an array of handlers that can then be filled externally. 

During the initialization of the PLIC (through the `plic_Init()` function the array is filled with
* The fixed handlers of the peripherals that are connected to x-HEEP (e.g. UART, SPI)
* a Dummy handler in any other case. 

---

# Changes

This change required several minor changes: 

* The `plic_get_irq_src_type` function was adapted to fill the handlers array. Now its called `plic_reset_handlers_list`. 
* The i2s HAL was updated to include a weak implementation of the handler, which only existed in the example before.
* The same logic that was implemented in the PLIC was also implemented for the GPIO so:
  * When there is a GPIO interrupt, no matter the ID, the PLIC will call the GPIO handler, passing the ID.
  * The GPIO handler will call a handler stored in the position of the ID relative to the first GPIO ID
  * These functions can be assigned from the application 

---

# Tests 

* The `example_dma` and `example_i2s` where tested on FPGA and worked. These are the only two HALs that include a handler right now. 
* The `example_gpio_intr` was tested on FPGA flash_load and flash_exec and worked. It implements two interrupts, with two different handlers.
* All apps were simulated in verilator and worked as expected. 

---

### :warning: Important

* The GPIO handler list depends on the fact that all GPIOs have consecutive IDs
* The I2c implementation of the interrupt in #265 need to be fixed (the substantiation of the first index moved to the I2C side)
* The future `sys_init()` function that will be developed by @JoseCalero will need to include `gpio_reset_all`, as there is no `gpio_init`. If that function is not called, then `gpio_reset_handlers_list` needs to be called. 
* 
---

# Future work

Some further changes might be needed

- [x] Many of the previous irq handlers where never overridden by the peripherals. They should be created. 
- [x] Right now its only possible to assign new handlers to the external interrupts, but maybe its a good idea to expand this possibility to, for instance, the gpios. 
- [x] Documentation
- [x] Discussion with @StefanoAlbini96


JuanSapriza:
The last changes were performed and the documentation in the first comment was updated. 
@StefanoAlbini96 @JoseCalero

JoseCalero:
i modified some minor things, but i still need to modify some other small things in here and I will merge it tomorrow

.gitignore
```diff
@@ -39,6 +39,7 @@ hw/system/pad_control/rtl/pad_control_reg_top.sv
 
 # same for the C header file and linker scripts
 sw/device/lib/runtime/core_v_mini_mcu.h
+sw/device/lib/runtime/core_intr.h
```
JoseCalero, line 42: why?

JuanSapriza: Good catch! This does not exist anymore. I will remove it and re-push

PR #368, opened 2023-08-28
Add check on `exec_ready_i` to `cv32e40x` decode stage

### LINKED ISSUE #365: `cv32e40x` bug: coprocessor issue information data (XIF) not sampled by the CPU
## Summary
The `cv32e40x` CPU attempts to offload an instruction to a coprocessor connected through the CORE-V eXtension Interface (XIF `issue_req`) whenever its instruction decoder fails to recognize it (or when the instruction is a CSR instruction). Currently, instruction offloading does *not* check for backpressure from the CPU execution stage, possibly resulting in the feedback information from the coprocessor (e.g., the intention to write back the offloaded instruction result to one of the CPU scalar GPRs `x0`-`x31`) not being correctly propagated to the ID-EX pipeline register and, therefore, to the following stages.

## Detailed example
The CPU execution stage applies backpressure to the instruction decoding stage (i.e., `ex_ready=0`) whenever execution takes more than one cycle, which is the case, for example, for memory accesses (`load`/`store` instructions). Take for instance the following assembly snippet:
```asm
...
li t0, 42
lw t1, 0(zero)
coproc.add t2, t1, t0
sw t2, 4(zero)
...
```
Where `coproc.add` is an (completely useless and redundant) instruction offloaded to a coprocessor that computes the sum `t0+t1` and send the result back to the CPU, to be stored in `t2`. `coproc.add` gets offloaded to the coprocessor as soon as it is decoded, in which cycle the previous `lw` instruction is being executed. Usually, the coprocessors instruction decoding stage can answer an offloading request as soon as it issued by the CPU, and the response data is retired the cycle after. In our example, the CPU execution stage is not ready to accept a new instruction when the `coproc.add` instruction is offloaded, becasue it is still processing the previous `lw`. Therefore, the coprocessor response data does *not* get sampled in the `ID-EX` pipeline register until the next cycle, where it is no longer valid, as shown in the following timing diagram:

issue

Information about the offloaded instruction that comes from the CPU IF-ID pipeline register gets sampled correctly, so the instruction is actually propagated through the CPU stages. However, crucial information in the coprocessor issue response, like the intention to write back data in the CPU GPR, or the possibility to raise exceptions, is not. One of the consequences is that the result data from the coprocessor does not get written back to the CPU GPRs. More sever consequences may appear when the coprocessor uses the CPU load-store unit or the bus, possibly triggering error or exceptions that would be ignored by the CPU.

## Proposed solution
A quick and dirty solution to get the information from the coprocessor correctly sampled into the ID-EX stage is to attempt instruction offloading through the XIF only when the execution stage is ready to accept a new instruction. This would result in the following fixed timing diagram:

solution

> NOTE: *this is not an optimal solution*, as it effectively stalls ID for one cycle. A better way would be to offload an instruction to the coprocessor as soon as it is fetched and decoded (as it is now), and sample the coprocessor response data in a dedicated register (different from the ID-EX pipeline register) while waiting for the execution stage to be ready. However, `cv32e40x` maintainers are certainly better-suited for defining and implementing a proper fix, so I would wait for their intervention on the core.

A pull request with the proposed fix implemented is being prepared and tested in [HEEPerator](https://eslgit.epfl.ch/blade/HEEPerator), and will be linked here soon.

StMiky:
## Description
Fix #365 by offloading instructions to a coprocessor only if the execution stage is ready to accept the new instruction. While this is a suboptimal solution, it prevents the ID-EX pipeline register from missing the XIF Issue Interface response from the coprocessor. More details are available in the associated issue.

This change is implemented as a patch to the `cv32e40x` core, while waiting for a better solution from its developers.

## Status
- [x] Test the proposed solution in [NM-Carus](https://eslgit.epfl.ch/blade/nm-carus)

### DIFF
--- a/hw/vendor/openhwgroup_cv32e40x/rtl/cv32e40x_id_stage.sv
+++ b/hw/vendor/openhwgroup_cv32e40x/rtl/cv32e40x_id_stage.sv
@@ -749,7 +749,9 @@ module cv32e40x_id_stage import cv32e40x_pkg::*;
       // Also attempt to offload any CSR instruction. The validity of such instructions are only
       // checked in the EX stage.
       // Instructions with deassert_we set to 1 from the controller bypass logic will not be attempted offloaded.
-      assign xif_issue_if.issue_valid     = instr_valid && (illegal_insn || csr_en) &&
+      // Only offload instructions if the EX stage is ready not to miss data from xif_issue.issue_resp
+      assign xif_issue_if.issue_valid     = instr_valid && ex_ready_i &&
+                                            (illegal_insn || csr_en) &&
                                             !(xif_accepted_q || xif_rejected_q || ctrl_byp_i.deassert_we);
 
       // Keep xif_offloading_o high after an offloaded instruction was accepted or rejected to get
--- a/hw/vendor/patches/openhwgroup_cv32e40x/0002-fix-xif-issue-backpressure.patch
+++ b/hw/vendor/patches/openhwgroup_cv32e40x/0002-fix-xif-issue-backpressure.patch
@@ -1,0 +1,15 @@
+diff --git a/rtl/cv32e40x_id_stage.sv b/rtl/cv32e40x_id_stage.sv
+index 1385dfb..11bc457 100644
+--- a/rtl/cv32e40x_id_stage.sv
++++ b/rtl/cv32e40x_id_stage.sv
+@@ -749,7 +749,9 @@ module cv32e40x_id_stage import cv32e40x_pkg::*;
+       // Also attempt to offload any CSR instruction. The validity of such instructions are only
+       // checked in the EX stage.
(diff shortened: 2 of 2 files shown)


davideschiavone:
@davidmallasen can you review it and see whether this may fix (or partially fix) your issue?

davidmallasen:
I tested this and it doesn't fix my issue. However, this doesn't seem to conflict with the other patch that we debugged to fix my issue and that I've been using since. This other patch is available here: https://github.com/davidmallasen/x-heep/blob/fix-cv32e40x/hw/vendor/patches/openhwgroup_cv32e40x/0002-Patch-fix-cv32e40x-xif.patch

I see the patch from this PR was included in a previous commit of X-HEEP. Could I create a PR also including my patch since they both seem to work together?

davideschiavone:
this is already done here: https://github.com/esl-epfl/x-heep/pull/392

PR #374, opened 2023-09-12
Readthedocs first integration

JoseCalero:
First integration of the documentation engine. Both for local building and for online readthedocs building management (you can already see the output in **[here](https://x-heep.readthedocs.io/en/latest/)**).
From this point we can already discuss how we want the documentation to be organized, displayed, etc.


JuanSapriza:
The procedure to make changes to this should be:
```bash
git remote add jose https://github.com/JoseCalero/x-heep
git pull jose readthedocs
git checkout -b readthedocs jose/readthedocs
# make your changes
git commit -am "I have made some changes"
git push -u jose readthedocs
cd docs
make html
```

JuanSapriza:
# :partying_face: Good news!

1. We now support markdown documents :)
2. We now can add documents in folders. 

As an example I added the `DMA.md` file in a `Peripherals` folder inside `docs/source`. 

## Requirements to add new documents

1. Add the new document in markdown to the `Peripherals` folder
2. Make sure the document has one single # header, otherwise they will be considered different documents. This is not bad per se... but we might want DMA to be a single point in the index, and not DMA, Introduction, Requirements, etc...
3. If a new folder is added, add it to the `toctree` inside `docs/source/index.rst` (as the `peripherals` folder is)
4. Commit and push
5. Open a terminal in the `docs` folder and run `make clean html`
6. Wait a few minutes and enjoy your brand new documentation in Read the Docs. 

## Additional changes

To accommodate all these changes I had to
1. Add a `.readthedocs.yaml` document in the root directory pointing to the configuration script and requirements
2. Include a new package as a requirement and in the config file to build markdown
3. Add a `toctree` at the end of the `index.rst` file (which I would have rather not)

JuanSapriza:
@JoseCalero One thing I noticed is that the docu is always seen until half the screen

image

Do you know if this is a configuration?

docs/source/How_to/ExternalDevices.md
```diff
@@ -92,8 +92,8 @@ To create and maintain a peripheral unit efficiently, use the `reggen` tool:
 
 2. Launch the `regtool.py` script to generate SystemVerilog RTL code and a C header file.
 
-For example, launching the script [`memcopy_periph_gen.sh`](./hw/ip_examples/memcopy_periph/memcopy_periph_gen.sh) generates 2 SystemVerilog files and one C header file:
```
JuanSapriza, line 95: All the references to the mempy peripheral should be removed

docs/source/How_to/GettingStarted.md
```diff
@@ -0,0 +1,123 @@
+# get started
```
JuanSapriza, line 1: The documentation inside the `How to` folder should be harmonized

JuanSapriza: We should also add a document on how to document.

JuanSapriza (approved):
There are some fixes that need to be done, but it can be merged imo

PR #385, opened 2023-09-28
fix GPIO HAL by adding selection of GPIO domain

davideschiavone:


davideschiavone:
@Ho3eintaji - what is the status of this?

PR #399, opened 2023-10-11
Update flash load linker with data_interleaved section

davideschiavone:


davideschiavone:
replacing #398

davideschiavone:
note to myself @davideschiavone,

So far the SPI examples fail (when flash_load is used) as they have been built under the assumption that LMA=VMA, 
whereas data in FLASH are not anymore at the same offset

We may need to use symbols from the linker script to get the offset displacement and add it to the application data addresses
to do the flash2memory copies

https://stackoverflow.com/questions/8398755/access-symbols-defined-in-the-linker-script-by-application

https://downloads.ti.com/docs/esd/SPRUI03/using-linker-symbols-in-c-c-applications-slau1318080.html

sw/applications/example_spi_flash_write/main.c
```diff
@@ -221,7 +232,9 @@ static inline __attribute__((always_inline)) void spi_config()
     spi_wait_for_ready(&spi_host);
 
     // Write command
-    const uint32_t write_byte_cmd = ((FLASH_ADDR << 8) | 0x02); // Program Page + addr
```
davideschiavone, line 224: I do not get why here the FLASH_ADDR is not reverted, @ruben-roalvarez ?

sw/device/lib/runtime/core_v_mini_mcu.c
```diff
@@ -0,0 +1,29 @@
...
+uint32_t * heep_get_data_address_lma(uint32_t* data_address_vma);
+uint32_t * heep_get_flash_address_offset(uint32_t* data_address_lma);
+
+// LINKER symbol, the value is defined in the LINKER script
+extern const uint32_t _lma_vma_data_offset;
+
+// this translates the virtual address to logical address for the FLASH
+uint32_t * heep_get_data_address_lma(uint32_t* data_address_vma){
+
+    uint32_t* data_address_lma = (uint32_t*) ((uint32_t)&(_lma_vma_data_offset) + (uint32_t)(data_address_vma));
```
JoseCalero, line 18: lma is load memory address (logical and virtual are the same).
question (maybe i am right now confused): _lma_vma_data_offset is already a memory offset, isn't it? i.e. you do not need to take the pointer to it (&), do you?

davideschiavone: true is load memory, I will change it - for what concern the offset, I thought the same but didnt work, let me try again maybe I made a mistake

davideschiavone: solved it in last commit

PR #401, opened 2023-10-16
Add `example_spi_host_quadIO` 

consanii:
Add `example_spi_host_quadIO` application in `sw/applications` subfolder.

It is based on `example_spi_host` and showcase the QuadSPI capabilities of the X-heep `spi_host` module.


sw/applications/example_spi_host_quadIO/main.c
```diff
@@ -0,0 +1,309 @@
...
+        .csaat      = true,              // Command not finished
+        .speed      = kSpiSpeedStandard, // Single speed
+        .direction  = kSpiDirTxOnly      // Write only
+    });
+    spi_set_command(&spi_host, cmd_read);
+    spi_wait_for_ready(&spi_host);
+
+
+    // Create segment 2
+    uint32_t read_byte_cmd = (REVERT_24b_ADDR(flash_original) | 0xFF << 24); // Fxh (here FFh) required by W25Q128JW
```
davideschiavone, line 221: why?

davideschiavone (changes requested):

sw/applications/example_spi_host_quadIO/main.c
consanii: [W25Q128JW datasheet](https://datasheet.lcsc.com/lcsc/2103261540_Winbond-Elec-W25Q128JWSIQ_C2763561.pdf) on page 33 describe the behaviour of line 221.
In short, after the command code at single speed the flash expect to receive the memory address MSB first at quad speed(so a Byte swap is require in a little-endian architecture). After it, a dummy byte at quad speed of value Fxh is required.
A detailed showcase of this behaviour can be also found in one [OpenTitan diagram](https://opentitan.org/book/hw/ip/spi_host/doc/programmers_guide.html#byte-ordering-conventions).

davideschiavone: sorry I was not clear, what I meant this comment `Fxh (here FFh) required by W25Q128JW` I do not understand what Fxh means

consanii: I found no clear description on the data sheet. I though it is an hexadecimal with first 4 bits all ones (F) and the next for can be whatever (x).
Anyway in the opentitan documentation (linked in the message before) I saw a diagram with the byte after the address with all ones so I decided to implement it like opentitan (FFh).

davideschiavone: ok thanks for the clarification

davideschiavone: another question @consanii , I see in the single mode SPI examples that the address is shifted by 8 after reversed, and then ored with `03`, here you do reverse, or with `ff`, but I do not see the shift by 8, why?

consanii: Because the behaviour in single and quad modes differs. In quad mode, the command code must be sent at standard speed before sending the address + Fxh afterward. There is no need to shift the address to make space for the command because the command is already transmitted.

davideschiavone:
- [ ] waiting for FPGA test

consanii:
QE bit is now set automatically by the application before read

Issue #411, opened 2023-10-18
App build failure in `run-helloworld` target


StMiky:
## Description
When calling the "all-in-one" (`run-helloworld` and `run-blinkfreertos`) targets on a clean clone of X-HEEP,  the software build process enters a recursive loop where infinite file descriptors are opened on the host system, eventually resulting in a build failure. Building the same applications with `make app PROJECT=...` works as expected.

## Proposed solution
Explicitly specifying the `SOURCE` when calling `make -C sw` in these tasks, like it's done in the `app` target, seems to resolve the issue. A pull request implementing this fix is being prepared.

## Host system
- OS: CentOS 8
- `cmake`: 3.20.2
- `gcc`: 12.1.1

JoseCalero:
should be fixed by #422

PR #412, opened 2023-10-18
Add explicit `SOURCE` to `run-helloworld` target

StMiky:
## Description
This change prevents the application build process from entering a recursive loop where infinite file descriptors are opened on the host system, eventually resulting in a build failure (see #411).

Only the "all-in-one" targets are affected (`run-helloworld` and `run-blinkfreertos`), and only when launched on a clean clone of X-HEEP. Building the same applications with `make app PROJECT=...` works as expected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,7 +165,7 @@ vcs-ams-sim:
 ## Uses verilator to simulate the HW model and run the FW
 ## UART Dumping in uart0.log to show recollected results
 run-helloworld: mcu-gen verilator-sim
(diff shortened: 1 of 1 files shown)


JuanSapriza:
Replicated the issue by
```bash
make clean-all
make verilator-sim
make  run-helloworld
# gets stuck
make app PROJECT=hello_world
# makes succefully
make run-helloworld
# makes and run successfully. 
```

However, after that, once I made `make run-helloworld` i get the full simulation model build and never again the issue. Therefore, i wouldn't be able to tell if the issue was solved in the PR :(
I did see after switching to the PR that WHEN running `make run-blinkyfreertos`, after the simulation model building I get an error that seems to be related to not getting the source. However, not adding the `SOURCE` should be an issue, because downstream the default path would be grabbed, if im not wrong. 

@StMiky is it possible that the fix worked for you because you had previously tried making `make app PROJECT=...`?

PR #417, opened 2023-10-19
expose DMA slots externally + external FIFO example

### LINKED ISSUE #407: expose the slot (fifo interface) of the DMA externally to HEEP for external streams accelerator/peripherals
@grinningmosfet

grinningmosfet:
Fix #407


hw/ip/i2s/rtl/i2s.sv
```diff
@@ -98,7 +98,6 @@ module i2s #(
   );
 
 
-
```
davideschiavone, line 101: please revert this change

hw/ip/i2s/data/i2s.hjson
```diff
@@ -109,7 +109,7 @@
       hwext:  "true"
       hwre:   "true"
       fields: [
-        { bits: "31:0", name: "RXDATA", desc: "latest rx data if DATA_READY flag is set", hwaccess: "hrw"}
```
davideschiavone, line 112: why?

hw/ip_examples/iffifo/rtl/iffifo.sv
```diff
@@ -0,0 +1,65 @@
...
+      .hw2reg,
+      .reg_req_i,
+      .reg_rsp_o,
+      .devmode_i(1'b0)
+  );
+
+  fifo_v3 #(
+      .DEPTH(DEPTH),
+      .DATA_WIDTH(WIDTH)
+  ) pdm2pcm_fifo_i (
```
davideschiavone, line 50: please change name here to iffifo_fifo_i

sw/device/lib/drivers/dma/dma.h
```diff
@@ -104,6 +104,8 @@ typedef enum
     DMA_TRIG_SLOT_SPI_FLASH_RX  = 4, /*!< Slot 3 (MEM < SPI FLASH). */
     DMA_TRIG_SLOT_SPI_FLASH_TX  = 8, /*!< Slot 4 (MEM > SPI FLASH). */
     DMA_TRIG_SLOT_I2S           = 16,/*!< Slot 5 (I2S). */
+    DMA_TRIG_SLOT_EXT0          = 32,/*!< Slot 6 (External peripherals 0). */
```
davideschiavone, line 107: please call them EXT_RX and EXT_TX

davideschiavone (changes requested):

hw/ip/i2s/data/i2s.hjson
grinningmosfet: I forgot to reset I2S files a modified. I'll revert this.
I did this to understand how the window-workaround worked.
(For information, it may be useless to add this.)

hw/ip/i2s/rtl/i2s.sv
grinningmosfet: I forgot to reset I2S files a modified. I'll revert this.

hw/ip_examples/iffifo/rtl/iffifo.sv
grinningmosfet: :+1:

sw/device/lib/drivers/dma/dma.h
grinningmosfet: :+1:

grinningmosfet:
PR converted to draft until status register is implemented.

grinningmosfet:
Interrupts are untested. External interrupts do not seem to work; i.e. in `external_dma_example`, `PRINTF("D");` in the `dma_intr_handler_trans_done` ISR is never executed.

davideschiavone:
> Interrupts are untested. External interrupts do not seem to work; i.e. in `external_dma_example`, `PRINTF("D");` in the `dma_intr_handler_trans_done` ISR is never executed.

@JoseCalero can help you with the SW part of the ISR

grinningmosfet:
PR converted to draft until interrupts are tested.

hw/fpga/xilinx_core_v_mini_mcu_wrapper.sv
```diff
@@ -191,7 +191,9 @@ module xilinx_core_v_mini_mcu_wrapper
       .pdm2pcm_pdm_io,
       .i2s_sck_io(i2s_sck_io),
       .i2s_ws_io(i2s_ws_io),
-      .i2s_sd_io(i2s_sd_io)
+      .i2s_sd_io(i2s_sd_io),
+      .ext_dma_slot_tx_i,
```
davideschiavone, line 195: this is wrong, it does not compile, plus, it should be 
`.ext_dma_slot_tx_i('0),`

and same with rx

davideschiavone (changes requested):

sw/applications/example_iffifo/main.c
```diff
@@ -0,0 +1,179 @@
...
+    if (ret != 0) {return EXIT_FAILURE;}
+    
+    if (compare_print_fifo_array() != 4) {return EXIT_FAILURE;}
+    
+    print_status_register();
+    
+    PRINTF("Launch MM -> Stream DMA\n");
+    dma_launch( &trans );
+    
+    int32_t read0 = mmio_region_read32(iffifo_base_addr, IFFIFO_FIFO_OUT_REG_OFFSET);
```
davideschiavone, line 136: if you are making these 2 read operations to empty the FIFO (pop operation) to let the DMA terminate, please add it in a comment

sw/applications/example_iffifo/main.c
```diff
@@ -0,0 +1,179 @@
...
+    trans.src        = &tgt_src;
+    trans.dst        = &tgt_dst;
+    trans.end        = DMA_TRANS_END_INTR;
+
+    ret = dma_validate_transaction( &trans, DMA_ENABLE_REALIGN, DMA_PERFORM_CHECKS_INTEGRITY );
+    if (ret != 0) {return EXIT_FAILURE;}
+    ret = dma_load_transaction(&trans);
+    if (ret != 0) {return EXIT_FAILURE;}
+    
+    if (compare_print_fifo_array() != 4) {return EXIT_FAILURE;}
```
davideschiavone, line 129: how do you compare "now" the 4 writes if the DMA did not terminate? would it be better to do the comparison only if the FIFO is full? if so, please add a `while(fifo_not_full())` before reading or something like this - do not rely on latency

sw/device/lib/drivers/dma/dma.h
```diff
@@ -104,6 +104,8 @@ typedef enum
     DMA_TRIG_SLOT_SPI_FLASH_RX  = 4, /*!< Slot 3 (MEM < SPI FLASH). */
     DMA_TRIG_SLOT_SPI_FLASH_TX  = 8, /*!< Slot 4 (MEM > SPI FLASH). */
     DMA_TRIG_SLOT_I2S           = 16,/*!< Slot 5 (I2S). */
+    DMA_TRIG_SLOT_EXT_TX            = 32,/*!< Slot 6 (External peripherals TX). */
```
davideschiavone, line 107: why the indentation here went wrong?

sw/device/lib/drivers/rv_plic/rv_plic.c
```diff
@@ -49,6 +49,7 @@
 #include "i2s.h"
 #include "dma.h"
 #include "spi_host.h"
+#include "iffifo.h"
```
davideschiavone, line 52: this should never be done, HEEP is used in other project without the testbench, so the plic functions should stay sane - move these parts in the application file

sw/device/lib/drivers/rv_plic/rv_plic.c
```diff
@@ -407,6 +408,10 @@ void plic_reset_handlers_list(void)
     {
       handlers[i] = &handler_irq_dma;
     }
+    else if ( i == EXT_INTR_1)
+    {
+      handlers[i] = &handler_irq_iffifo;
```
davideschiavone, line 413: as above, this should never be done, HEEP is used in other projects without the testbench, so the plic functions should stay sane - move these parts in the application file - for example by exploiting the irq_dummy or something like this

davideschiavone (changes requested):
I left comments in the code to be addressed before merging

grinningmosfet:
PR converted to draft change request is addressed.

sw/applications/example_iffifo/main.c
grinningmosfet: This is the first `compare_print_fifo_array()`, I call it once before both transactions to make sure the arrays do not match. The second call is placed after the `protected_wait_for_dma_interrupt()` method that makes sure both transactions are terminated before comparison. I need to update `protected_wait_for_dma_interrupt()` though.

sw/device/lib/drivers/rv_plic/rv_plic.c
grinningmosfet: Switched to `plic_assign_external_irq_handler` implementation instead.

sw/device/lib/drivers/rv_plic/rv_plic.c
grinningmosfet: Switched to `plic_assign_external_irq_handler` implementation instead.

sw/applications/example_iffifo/main.c
davideschiavone: before reading the fifo, can you do "while(fifo_full == 0);" so that you read only if the FIFO is full? of course you have to do it twice

davideschiavone (changes requested):

sw/applications/example_iffifo/main.c
davideschiavone: actually you do not need that as the reading from the FIFO should be blocking in case it's empty, right?

sw/applications/example_iffifo/main.c
```diff
@@ -0,0 +1,187 @@
...
+  PRINTF("}\n");
+  return errors;
+}
+
+void print_status_register(void)
+{
+  mmio_region_t iffifo_base_addr = mmio_region_from_addr((uintptr_t)IFFIFO_START_ADDRESS);
+  int32_t status = mmio_region_read32(iffifo_base_addr, IFFIFO_STATUS_REG_OFFSET);
+  PRINTF("STATUS = ");
+  PRINTF(status & 1 ? "E" : "-"); // FIFO empty
```
davideschiavone, line 83: pls use the following defines avilable in the regs.h 

#define IFFIFO_STATUS_EMPTY_BIT 0
#define IFFIFO_STATUS_AVAILABLE_BIT 1
#define IFFIFO_STATUS_REACHED_BIT 2
#define IFFIFO_STATUS_FULL_BIT 3

sw/applications/example_iffifo/main.c
davideschiavone: as the fifo is not blocking, reading to an empty fifo succeeds, so the requirements about reading if fiflo_full is strong, but could also be "if not empty"

grinningmosfet: Missed that, yes, totally makes sense! I'll put `while(fifo_full == 0);` as I want the DMA to be blocked.

PR #419, opened 2023-10-24
Add stack and heap size as parameters to mcu-gen

Luigi2898:
## Description
To run bigger applications, it can be useful to configure heap and stack to have a bigger dimension than the default one.
To manually modify the linker script or the template with a different value is inconvenient, so two parameters are added to the linker script template in order to set this value with the mcu_cfg file.

## Proposed solution
`link.ld.tpl` template is modified to read this parameter and `mcu_gen.py` is modified accordingly.
The default `mcu_cfg.hjson` file is modified to guarantee that the default value is the same as before.

### DIFF
--- a/mcu_cfg.hjson
+++ b/mcu_cfg.hjson
@@ -27,6 +27,8 @@
                 lenght: whatisleft, #keyword used to calculate the size as: ram.length - code.lenght
             }
         },
+        stack_size: 0x800,
+        heap_size: 0x800,
     }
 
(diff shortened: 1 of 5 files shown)


davideschiavone:
Thanks @Luigi2898 , can you also add it to the others linker scripts?

Luigi2898:
@davideschiavone , thank you for your answer. I added it with the last commit.

PR #422, opened 2023-10-27
Compilation fix

JoseCalero:
This fix is related to closed PR #412 and issue #411. 
So, we were not exporting properly the ENV variables to sub-makefiles, i.e. to children makefiles, and so the SOURCE was not propagated properly. This led to problems when executing from scratch, i.e. not having still nothing into the CMake cache.


StMiky:
Hi @JoseCalero, I've just tested the PR on a clean clone of your fork (no previous `make app` or anything). It compiles and runs the simulation flawlessly, so I'd say it's solved. Thank you!

PR #424, opened 2023-10-31
Fix run-blinky-freertos command

JoseCalero:


JoseCalero:
@StMiky this was the problem, the name was wrong, as recently we changed all the app names. That is why it was not founding those files. In fact, they do not exist :)

PR #426, opened 2023-11-07
Add standard and quad write functionality to flash model

consanii:
Add standard and quad write capabilities to the current verilog flash model. This allows write functions to be executed also on simulation.

This pull request also includes a write example application that was used during test phase of the newly added flash functionalities.


sw/applications/example_spi_flash_write/main.c
```diff
@@ -60,7 +60,7 @@
 
 /* By default, printfs are activated for FPGA and disabled for simulation. */
 #define PRINTF_IN_FPGA  1
-#define PRINTF_IN_SIM   0
+#define PRINTF_IN_SIM   1
```
davideschiavone, line 63: please put it back to 0

sw/applications/example_spi_flash_write/main.c
```diff
@@ -367,12 +360,15 @@ int main(int argc, char *argv[])
     }
     PRINTF("triggered\n\r");
 
+    // In simulation there is no need to wait
+    #ifdef USE_SPI_FLASH
```
davideschiavone, line 364: to avoid too many ifdef, please let's wait in simulation as well, can you please revert this change?

sw/applications/example_spi_flash_write_quad/main.c
```diff
@@ -0,0 +1,446 @@
...
+        .csnlead    = 0xF,
+        .fullcyc    = false,
+        .cpha       = 0,
+        .cpol       = 0
+    });
+    spi_set_configopts(&spi_host, 0, chip_cfg);
+    spi_set_csid(&spi_host, 0);
+
+    // Reset
+    const uint32_t reset_cmd = 0xFFFFFFFF; // ???
```
davideschiavone, line 154: @consanii , can we replace this command with the real reset please? as @ruben-roalvarez said it is wrong

sw/applications/example_spi_flash_write_quad/main.c
```diff
@@ -0,0 +1,446 @@
...
+        .csaat      = false, 
+        .speed      = kSpiSpeedQuad, // QUAD SPEED 
+        .direction  = kSpiDirTxOnly
+    });
+    spi_set_command(&spi_host, cmd_write1);
+    spi_wait_for_ready(&spi_host);
+
+    // Wait for all the fifo to be drained
+    spi_wait_for_tx_empty(&spi_host);
+    #ifndef TARGET_SIM
```
davideschiavone, line 274: let's wait for response also in simulation please

davideschiavone (changes requested):

sw/applications/example_spi_flash_write_quad/main.c
ruben-roalvarez: I can double-check the datasheet in any case, but maybe @consanii you can do three tests: write the FLASH without a reset, with all Fs reset, and with the reset you found in the datasheet (all without previously sending any command from the PC). Let's see if it fails for any of the cases

sw/applications/example_spi_flash_write_quad/main.c
ruben-roalvarez: It will only work in questa, right?

consanii: Same problem, the function will never return.

sw/applications/example_spi_flash_write_quad/main.c
consanii: I will do the test but I think it will work in both three cases, as the reset is not needed in the first place. Moreover, even if the reset was needed, it would be redundant as the reset command revert the board to the initial power-on state. And the board is already in the initial power on state before the write command (as it's just turned on).

consanii:
I added the requested modifications.

sw/applications/example_spi_flash_write_quad/main.c
ruben-roalvarez: Amazing! Thanks!

PR #439, opened 2024-01-12
fix memset bug

consanii:
As title, fix `memset()` function in `example_sp_readi` test not being called.


davideschiavone:
hi @consanii , you first need to rebase your branch to the X-HEEP main branch, otherwise as you see there are conflicts

consanii:
The only PROBLEM that I see is on the `example_spi_read` file which of course as conflicts on where I modified it.

davideschiavone:
it is because you have to merge in your branch the main, otherwise just create a new branch aligned to the main branch of HEEP and apply there your commit - github sees that you are applying the changes that in this PR are listed into Files changed - which is not the case

image

consanii:
I merged main, and now everything looks ok. It just still shows all the commits, and I'm not getting why.

PR #444, opened 2024-02-09
fix compilation errrors reported by xcelium

leojz:
there are some features that xcelium reports as not supported, others as illegal syntax.
With these updates x-heep successfully compiles rtl and testbench.

### DIFF
--- a/hw/ip/i2s/rtl/i2s_core.sv
+++ b/hw/ip/i2s/rtl/i2s_core.sv
@@ -7,8 +7,8 @@
 // Description: I2s core logic 
 
 module i2s_core #(
-    parameter MaxWordWidth,
-    parameter ClkDividerWidth
+    parameter MaxWordWidth = 32,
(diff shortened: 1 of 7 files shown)


leojz:
Hi @davideschiavone ,

I'm a still a beginner with git and the concept of vendoring is new to me. Please go ahead with your suggestion, I'll try to follow.

There is one additional change required for a successful compilation; An update for vendor patch lowrisc_opentitan/uartdpi.patch. Don't know how to accomplish this. The \`ifndef VCS +MODELSIM need to be complemented with one for XCELIUM. An alternative might be to replace the former with `ifdef VERILATOR. What do you think?

Updates for .core and Makefile are ready.  

Cheers, Leo

davideschiavone:
Hi @leojz can you please push the Makefile and .core ?

davideschiavone:
replaced in #452

PR #466, opened 2024-03-05
Structs multireg fix

StefanoAlbini96:
Fixed a bug in `structs_gen.py` regarding the generation of multiregs.
Some registers were not created properly due to a wrong integer approximation.


JoseCalero (approved):
approved

PR #472, opened 2024-03-15
Update cve2 to latest commit

Luigi2898:
This PR is to update the cve2 core to the latest verion in https://github.com/openhwgroup/cve2


Luigi2898:
@davideschiavone
I tested it with modelsim running the `hello_world`, the issue was at build time, so the `hello_world` should be enough. 
Verilator simulation of all the apps is done in CI, so that should be fine too.

This PR is now ready for review.

Issue #468, opened 2024-03-06
MATCHES statements in CMakeLists.txt may lead to false positives and cause the wrong target to be built


StMiky:
## Description
The `MATCHES` statements in `sw/CMakeLists` succeed with any absolute file path containing the word assigned to `$TARGET`, even before the `sw/device/target/` part of the path. When this happens, multiple include paths for `x-heep.h` are passed to the compiler, causing the wrong headers to be used in the build process.

## Example
Assume a poor X-HEEP user is called _Simon_. Simon's user name on the server where he cloned the X-HEEP repository is `simon.hellodarkness`. Therefore, each file in the lists defined by `cmake` will contain the substring `/home/simon.hellodarkness/<path-to-xheep>/` in its full path, and therefore the `sim` string. If Simon decides to build an app with the default target `TARGET=sim`, the `MATCHES` statement at [this line](https://github.com/esl-epfl/x-heep/blob/01c514c19ef95b050a7f77141435c452a8242bd2/sw/CMakeLists.txt#L68) of `CMakeLists.txt` will match all his files in `/home/simon.hellodarkness/<path-to-xheep>/sw/*`, including all three `x-heep.h` headers from `sw/target/nexys-a7-100t/`, `sw/target/pynq-z2/` and `sw/target/sim/`. Because `nexys-a7-100t` comes first in alphabetical order, the FPGA `x-heep.h` is used in the build process in place of the one in `target/sim/`.

### Reproduce
```bash
mkdir poor-simon && cd poor-simon
git clone https://github.com/esl-epfl/x-heep.git
cd x-heep
make mcu-gen
make app PROJECT=hello_world TARGET=sim # note the pragma messages reporting the wrong header being used
```

## Solution
Because there is a remote possibility that Simon refuses to change his name, it is probably better to implement stricter match patterns in `CMakeLists.txt`. A pull request with a tentative solution is already being prepared.

JuanSapriza:
I think you make a very relevant point about the level of commitment we wxpect from our X-HEEP users. Maybe we should review that policy hahaha 

I think this should be solvable through #469

PR #473, opened 2024-03-16
Fix cmake error

Luigi2898:
# Description

If X-HEEP is vendorized and software is built as described in the [docs](https://github.com/esl-epfl/x-heep/blob/main/docs/source/How_to/eXtendingHEEP.md#building-software) the compilation fails.

## Possible issue

The check `(${file_path} MATCHES "${ROOT_PROJECT}device/")` does not match any header file if the `ROOT_PROJECT` variable is set to the vendorizing directory.

## Solution

Make the check `(${file_path} MATCHES "${SOURCE_PATH}device/")` so that symllinks are matched correctly.

:warning: @StMiky check with users whose name matches with the regex  `/\b\w*(?:sim|Sim)\w*\b/gm` that this solution does not break the solution to [your issue](https://github.com/esl-epfl/x-heep/issues/468).


JoseCalero:
Thanks @Luigi2898 for spotting this problem for vendorize projects. Indeed, this was modified due to user naming problems (@StMiky), but it caused problems when carrying the path for vendorized roots. I will check it on Monday, one possible solution is to avoid having any `sim` into the project and so we can have `simulation`, i hope nobody is called `simulation` :D

StMiky:
Thank you @Luigi2898 for spotting the wrong variable (`ROOT_PROJECT` instead of `SOURCE_PATH`. I didn't look into `sw/makefile` when I proposed that solution.

Luigi2898:
@StMiky I tested it in my application after your commits and it works fine! Thank you.

StMiky:
In the meantime, I also fixed a similar issue with file paths containing the word `${PROJECT}`, that made `MATCHES` statements like [this](https://github.com/esl-epfl/x-heep/blob/76d58efe7b9dec0723c1cb9aaf8ad76ad6c85373/sw/CMakeLists.txt#L131) fetch files from outside the X-HEEP repository.

For example, having X-HEEP inside a directory containing `minver`, like `minver-accelerator/hw/vendor/x-heep/` resulted in the build process for the app `minver` fetching sources from the entire `minver-accelerator` directory tree.

JoseCalero (approved):
Approved

JoseCalero:
Thanks @StMiky ! All changes make sense. I just did a quick test with X-HEEP main repo (not vendorized) and it worked. As @Luigi2898 also tried it with one vendorized and also worked, I will merge. Actually, yesterday we spotted the same problem with heepsilon (another xheep vendorized version), so this is also fixing that! (Anyway on Monday, Majo will be re-testing that as well)

PR #476, opened 2024-03-27
DSP basic_maths_functions

Antoine-V74:
To the attention of Dr. José Miranda : 
I tried and simulate every basic maths functions of the pulpino and they gave the correct results.
To use them the makefile needs to be changed, as it does no recognize the riscv_maths file


sw/applications/example_dsp/riscv_sub_q7.c
```diff
@@ -0,0 +1,130 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include "/home/antoine/FORKs/CMSIS-DSP-PULPino/inc/riscv_math.h"
```
davideschiavone, line 3: you should not use absolute path

sw/applications/example_dsp/riscv_sub_q7.c
```diff
@@ -0,0 +1,130 @@
+#include <stdio.h>
```
davideschiavone, line 1: missing license -

davideschiavone (changes requested):

davideschiavone:
@JoseCalero - can we close this?

PR #479, opened 2024-04-09
Add core_v_mini_mcu.sv to gitignore

davidmallasen:
This file is generated with the `core_v_mini_mcu.sv`, but it is not in the gitignore


davideschiavone:
this is done on purpose to keep the files readable from github GUI, but it will change in a current PR

PR #488, opened 2024-04-20
split fpga .core

davideschiavone:
@LuigiGiuffrida98 , can you pls test it?


LuigiGiuffrida98:
@davideschiavone I tested it on the pynq-z2 and it seems to work.

PR #491, opened 2024-04-22
Remove dependency from `bc`

cousteaulecommandant:
`bc` is used in the Makefile to perform simple arithmetic.  While this utility is likely to be installed, it makes more sense to just use the shell's native `$((...))` feature.


davidmallasen:
Is this needed in any environment @cousteaulecommandant? Otherwise, I would prefer to err on the side of caution and not merge to avoid breaking any setups.

cousteaulecommandant:
If I recall correctly, the setup I used was based on a minimal Ubuntu setup inside a docker container and it didn't come with bc by default so updating to a newer version of X-HEEP broke the flow.

Overall, I don't think it makes sense to depend on an external application (`bc`) that may not be installed, when the standard shell already does that (and saves you one process) via `$(( ))`.

davidmallasen:
Could you rebase to the develop branch? We'll stage the change there and merge to main later

cousteaulecommandant:
Rebased.

PR #493, opened 2024-04-23
extend mcu gen pads

davidmallasen:
Extend the pad configuration so it includes an optional mapping to the side of the die. An example of this extra configuration is as follows:

~~~
module pad_cell_inout #(
    parameter int unsigned PADATTR = 16,
    parameter int unsigned SIDE = 0, // 0: top, 1: right, 2: bottom, 3: left
) (
    input logic pad_in_i,
    input logic pad_oe_i,
    output logic pad_out_o,
    inout logic pad_io,
    input logic [PADATTR-1:0] pad_attributes_i
);
    if (SIDE == 0) begin
        // Instantiate the top mapping pad cell
    end else if (SIDE == 1) begin
        // Instantiate the right mapping pad cell
    end else if
    // ...
endmodule

~~~

### DIFF
--- a/hw/core-v-mini-mcu/include/core_v_mini_mcu_pkg.sv.tpl
+++ b/hw/core-v-mini-mcu/include/core_v_mini_mcu_pkg.sv.tpl
@@ -190,5 +190,11 @@ package core_v_mini_mcu_pkg;
 
   localparam int unsigned NUM_PAD_PORT_SEL_WIDTH = NUM_PAD > 1 ? $clog2(NUM_PAD) : 32'd1;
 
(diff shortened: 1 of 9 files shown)


davideschiavone:
On top of adding the enum for orientarion in the script, can you also add it in the core_v_mini_mcu_pkg so that the pad design can rely on such SV enum rather than hardwiring 1 2 3 4?

davidmallasen:
Done @davideschiavone . Should I modify the ones in `hw/simulation/` or in `hw/fpga/`? I see the ones for simulation have an extra parameter that the ones for fpga don't have

PR #498, opened 2024-04-26
extend the sign of the dest data in DMA

LuigiGiuffrida98:
This PR is to implement sign extension capability in the DMA data transfer if the data type of the destination is wider than the data type of the source.

This required the addition of little logic in the DMA and three control registers:
- DST_DATATYPE
- SRC_DATATYPE
- SIGN_EXTEND

TODO:

- [x] Edit RTL
- [x] Add example
- [x] Edit HAL
- [x] Add HAL to the example
- [x] Test with verilator are passing
- [x] Test on pynq-z2 are passing
     - [x] example_dma
     - [x] example_dma_2d
     - [x] example_spi_read
     - [x] example_spi_write


LuigiGiuffrida98:
@JuanSapriza, I modified the DMA HAL and the example_dma application, can you please review them?

JuanSapriza:
To simulate `example_dma` we need mora than 2 memory banks (the default). We should do something about that or the tests are going to fail. How was it being done before this? :thinking:

JuanSapriza:
Apart from that, all tests passed. I will check the rest of the code for blasphemies

JuanSapriza:
I took the liberty of changing a few words, but everything looks amazing. 
Also tested `example_dma_2d` and `example_dma_external` and both return `SUCCESS`. 

:heavy_check_mark: Good to merge on my side

PR #505, opened 2024-05-08
Make the Makefile help extendible

JuanSapriza:
To allow projects vendorizing X-HEEP to use the Makefile target documentation generation (what is printed when you run only `make`) I needed to make a small change in the Makefile. 

I also removed an unnecessary intermediate file and updated the documentation so we can replicate this on all other projects. 

I tested this [on a HEEPsilon branch](https://github.com/JuanSapriza/HEEPsilon/tree/makefile_automatic_generation).


JuanSapriza:
@JoseCalero This is a request for HEEPatia

LuigiGiuffrida98:
I tested it too and it works beautifully

PR #511, opened 2024-05-15
Moves peripheral to new configuration script

mBornand:
HW part should be mostly ready.
SW is work in progress.

The configuration that works:
`make mcu-gen X_HEEP_CFG=configs/example.py  && make verilator-sim`

The python environment needs to be recreated for this to work.


mBornand:
With the latest changes at least hello wold runs.

PR #512, opened 2024-05-16
Dynamic memory allocation issue

LinusCrugnola:
- Example of malloc not working as intended in first commit
- A quick fix for some of the issues in the second commit
---
Remaining problems:
- NULL suddenly get 0x8
- If only checking malloc() return for not being NULL, it still does not work


davideschiavone:
@jmiranda - can we close this?

PR #516, opened 2024-05-22
Fix example_gpio_intr

mBornand:
ensure the counter can restart

If we do not do this the application freezes because the counter never resets.


davideschiavone:
@simone-machetti , can you please review it?

Issue #519, opened 2024-05-29
Including assembly files in compilation


JuanSapriza:
> to test the dual core configuration, I need to compile a .S file (assembly) in my application, is there a way to enhance the CMAKE flow to not only include the C files but also the ASM? it should be quite easy, I did smth similar here: https://github.com/esl-epfl/x-heep/pull/449/files
> but there it was among the libraries, here I would like you to find for any ".S" file in the PROJECT folder and include it in the compilation flow - can you pls guys do it on the main heep so that I can continue with the dual core?

Davide Schiavone, circa 2024

@JuanSapriza @davideschiavone @JoseCalero

davideschiavone:
@Ho3eintaji as well

JuanSapriza:
Fixed in #522

PR #522, opened 2024-06-03
Include assembly files as source files for compilation

JuanSapriza:
As a request of #519 I have added the inclusion of assembly (.s) files into the linked files. 

@davideschiavone @Ho3eintaji do you have a test example?

### DIFF
--- a/sw/applications/example_asm/add.s
+++ b/sw/applications/example_asm/add.s
@@ -1,0 +1,6 @@
+    .section .text
+    .globl add
+
+add:
+    add a0, a0, a1   # Add the values in a0 and a1, store the result in a0
+    ret              # Return from the function
--- a/sw/applications/example_asm/add.s
+++ b/sw/applications/example_asm/add.s
@@ -1,6 +1,6 @@
     .section .text
-    .globl add
+    .globl add_asm_function
 
-add:
+add_asm_function:
     add a0, a0, a1   # Add the values in a0 and a1, store the result in a0
     ret              # Return from the function
--- a/sw/applications/example_asm/main.c
+++ b/sw/applications/example_asm/main.c
@@ -1,4 +1,5 @@
 #include <stdio.h>
+#include <stdlib.h>
 #include "constants.h"
 #include "x-heep.h"
 
@@ -27,5 +28,5 @@ int main() {
     PRINTF("%d+%d=%d\n", num1, num2, sum);
     PRINTF("%d*%d=%d\n", num2, MULTIPLY_CONSTANT, mul );
     
-    return EXIT_SUCCESS;   
+    return (sum == num1+num2) && (mul == num2*MULTIPLY_CONSTANT) ? EXIT_SUCCESS : EXIT_FAILURE;   
 }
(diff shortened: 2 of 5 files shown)


sw/applications/example_asm/add.s
```diff
@@ -0,0 +1,6 @@
+    .section .text
+    .globl add
+
+add:
```
davideschiavone, line 4: can you please call it "add_asm_function" ? easier to find it in the object dump

sw/applications/example_asm/main.c
```diff
@@ -0,0 +1,13 @@
...
+
+// Declaration of the assembly function for the little Juan
+extern int add(int a, int b);
+
+int main() {
+    int num1 = 10;
+    int num2 = 20;
+    int result = add(num1, num2);
+
+    printf("%d+%d=%d\n", num1, num2, result);
```
davideschiavone, line 11: return result!=30 and make printf --> PRINTF with ifdef

davideschiavone (changes requested):

JuanSapriza:
Added compatibility with both `.s` and `.S` files. For this I needed to exclude the `.S` files contained inside the `device` directory, or would have collisions with `crt` and so on. 

I have tested this compiling for verilator (+ the function returns successfully). 
```bash
make run-app-verilator PROJECT=example_asm
```

davideschiavone (approved):
LGTM, let's wait for @JoseCalero

sw/applications/example_asm/main.c
```diff
@@ -0,0 +1,32 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include "constants.h"
+#include "x-heep.h"
+
+/* By default, printfs are activated for FPGA and disabled for simulation. */
+#define PRINTF_IN_FPGA  1
+#define PRINTF_IN_SIM   0
+
+#if TARGET_SIM && PRINTF_IN_SIM
```
JoseCalero, line 10: `#if TARGET_SIM && PRINTF_IN_SIM
#elif PRINTF_IN_FPGA
#else
    #define PRINTF(...)
#endif`

sw/applications/example_asm/main.c
```diff
@@ -0,0 +1,32 @@
...
+        #define PRINTF(fmt, ...)    printf(fmt, ## __VA_ARGS__)
+#elif TARGET_PYNQ_Z2 && PRINTF_IN_FPGA
+    #define PRINTF(fmt, ...)    printf(fmt, ## __VA_ARGS__)
+#else
+    #define PRINTF(...)
+#endif
+
+// Declaration of the assembly function for the little Juan
```
JoseCalero, line 18: can you remove the 'little Juan' thing? (i guess that is my fault xD)

JoseCalero (changes requested):
please, resolve the two comments i left

sw/applications/example_asm/main.c
JoseCalero: yes, we are porting more and more boards, so actually adding just the pynq is not OK, you should be adding also the zcu and the nexys. To avoid that (for now...) let's leave it as general as the sniped i pasted

PR #525, opened 2024-06-10
extend riscv_dbg to support multi Harts

LuisDonatien:
Fixed some bugs in pulp_platform_riscv_dbg to allow configuration of multiple harts.

Added lint_off rule to avoid warning.

### DIFF
--- a/hw/core-v-mini-mcu/core_v_mini_mcu.sv
+++ b/hw/core-v-mini-mcu/core_v_mini_mcu.sv
@@ -11,10 +11,12 @@ module core_v_mini_mcu
     parameter ZFINX = 0,
     parameter EXT_XBAR_NMASTER = 0,
     parameter X_EXT = 0,  // eXtension interface in cv32e40x
+    parameter EXT_HARTS = 0,
     //do not touch these parameters
     parameter EXT_XBAR_NMASTER_RND = EXT_XBAR_NMASTER == 0 ? 1 : EXT_XBAR_NMASTER,
     parameter EXT_DOMAINS_RND = core_v_mini_mcu_pkg::EXTERNAL_DOMAINS == 0 ? 1 : core_v_mini_mcu_pkg::EXTERNAL_DOMAINS,
-    parameter NEXT_INT_RND = core_v_mini_mcu_pkg::NEXT_INT == 0 ? 1 : core_v_mini_mcu_pkg::NEXT_INT
+    parameter NEXT_INT_RND = core_v_mini_mcu_pkg::NEXT_INT == 0 ? 1 : core_v_mini_mcu_pkg::NEXT_INT,
+    parameter EXT_HARTS_RND = EXT_HARTS == 0 ? 1 : EXT_HARTS
 ) (
 
     input logic rst_ni,
@@ -300,6 +302,8 @@ module core_v_mini_mcu
     output reg_req_t ext_peripheral_slave_req_o,
     input  reg_rsp_t ext_peripheral_slave_resp_i,
 
+    output logic [EXT_HARTS_RND-1:0] ext_debug_req_o,
+
     input logic [NEXT_INT_RND-1:0] intr_vector_ext_i,
 
     output logic cpu_subsystem_powergate_switch_no,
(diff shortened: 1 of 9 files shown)


davideschiavone:
Hello @LuisDonatien , thanks for your fixes.
It seems that you are fixing the pulp debugger, so you should send your PR to this repository https://github.com/pulp-platform/riscv-dbg maintained by ETH and University of Bologna.

In case the fix was already there, then you should just update the vendor hash commit and revendorize the debugger

LuisDonatien:
> Hello @LuisDonatien , thanks for your fixes. It seems that you are fixing the pulp debugger, so you should send your PR to this repository https://github.com/pulp-platform/riscv-dbg maintained by ETH and University of Bologna.
> 
> In case the fix was already there, then you should just update the vendor hash commit and revendorize the debugger

Hello @davideschiavone, thanks for the quick response.
I'm going to do the PR to the pulp repo and update the hash when it will be accepted.
I will also update this PR with another commit involving the debug_subsystem.sv to allow multiple harts as well.

davideschiavone:
Thanks @LuisDonatien , jn the meantime feel free to update the debug subsystem in this PR, you don't need another separate PR in my opinion

LuisDonatien:
Implemented hartinfo_t initialization vector for multiple hart configuration.

davideschiavone:
thanks @LuisDonatien , if you patch the pulp debug I can merge it in after testing it - only one remark, NHART should be exposed at the very top level of core_v_mini_mcu, otherwise none can change this parameter - it is fine to not expose it to X-HEEP as X-HEEP itself has one HART by default, and when you extend it you usually use core_v_mini_mcu - can you also do that pls? 

@danivz can help you patch the pulp debugger, or if any of the 2 can do it - reach out to me by email

danivz:
Patch for vendor done!

LuisDonatien:
Updated NRHARTS exposed at the top level of core_v_mini_mcu through the EXT_HARTS parameter.

Done.

davideschiavone:
thanks a lot - it looks great - I will have time next week to test it but most probably I am gonna merge as it is

hw/core-v-mini-mcu/core_v_mini_mcu.sv
```diff
@@ -488,7 +494,7 @@ module core_v_mini_mcu
       .jtag_trst_ni,
       .jtag_tdi_i,
       .jtag_tdo_o,
-      .debug_core_req_o(debug_core_req),
+      .debug_core_req_o(debug_req),
```
davideschiavone, line 497: hi @LuisDonatien , can we also pls expose the debug_reset_n signal? I am not sure but I think there is one for each hart as well

davideschiavone (changes requested):

hw/core-v-mini-mcu/core_v_mini_mcu.sv
LuisDonatien: Hi Davide.

I see that your debug_reset_n signal comes from the ndmreset_o in the dm_obi_top inside the debug_subsystem module.
It is more like a general reset for the whole system except for the debugger itself. I think you are right that it has to be exposed externally as well.

**Additional proposal**
In addition to debug_reset_n signal it would be possible to expose the cpu_subsystem_rst_n signal externally as ext_cpu_subsystem_rst to ensure that external harts access peripherals and bus_system correctly initialized?

I can do the PR and you will review the changes.

davideschiavone:
@LuisDonatien , I think we need to change also the FPGA wrapper https://github.com/esl-epfl/x-heep/blob/main/hw/fpga/xilinx_core_v_mini_mcu_wrapper.sv otherwise we break the FPGA flow

LuisDonatien:
> @LuisDonatien , I think we need to change also the FPGA wrapper https://github.com/esl-epfl/x-heep/blob/main/hw/fpga/xilinx_core_v_mini_mcu_wrapper.sv otherwise we break the FPGA flow

@davideschiavone 
Do you mean to expose ext_harts parameters through xheep_system to fpga wrapper parameters?

davideschiavone:
> > @LuisDonatien , I think we need to change also the FPGA wrapper https://github.com/esl-epfl/x-heep/blob/main/hw/fpga/xilinx_core_v_mini_mcu_wrapper.sv otherwise we break the FPGA flow
> 
> @davideschiavone Do you mean to expose ext_harts parameters through xheep_system to fpga wrapper parameters?

I mean to add the new output pins in the xheep instance of the vivado wrapper, otherwise vivado complains about missing pins

Issue #494, opened 2024-04-23
Add target information to `pad_cfg.hjson` to centralize pad information


davidmallasen:
In `pad_cfg.hjson` we list the pads, but there is no information on what is the target where they will be placed or used (pynq-z2, nexys, zcu104, ASIC...). Together with @christophmuellerorg we discussed about centralizing all of this information in this file to generate automatically the pads and make sure this information stays consistent throughout the different places where it is used. For the ASIC targets this would have the mapping of the cells (top, right, bottom...) and other information needed for their placement. This would move the extra config info added in #493 . For the FPGA targets it would have all of the pin assignments so they can be modified externally without having to apply patches. An example of the `targets:`  that would be placed alongside the `pads:` is the following:

~~~
targets: {
        asic_pkgname_vendor_node {
            type: wirebond,
            package: pkgname,
            vendor: tsmc,
            node: 0,
            mapping: {
                top: {
                    clk: {
                        cell: "",  // either type or cell have to be specified
                        pin_name: auto  // optional
                        layout_pin_number: -1,  // optional
                        package_pin_number: -1,  // optional
                        bond_pad: true,  // optional
                        extra_parameters: "",  // optional
                        bondpad_extra_parameters: ""  // optional
                    }
                },
                top_right: {
                    corner_tr: {
                        num: 1,
                        cell: corner_cell_name,
                        orientation: // One of R90, R180, R270...
                    }
                },
                right: {

                },
                bottom_right: {

                },
                bottom: {

                },
                bottom_left: {

                },
                left: {

                },
                top_left: {

                }
            }
        }

        pynq-z2 {
            type: fpga,
            vendor: xilinx,
            part: xc7z020clg400-1,
            mapping: {
                clk: {
                    pin: A1,
                    IO_std: LVCMOS33
                },
                // TBD
            },
            additional_constraints: {

            }
        }
    }
~~~

PR #526, opened 2024-06-12
Add target info to the configuration of the pads

davidmallasen:
See #494.
- [x] Move the mapping to the target
- [x] Add support for the rest of the target configurations


davidmallasen:
@davideschiavone @christophmuellerorg Can you review this?
Running mcu-gen will generate the `hw/asic/generic_example/pnr/inputs/test.io` so you can test.
If there are any changes that I should make, please let me know. The behavior now should be analogous to that of the python script you told me.

PR #527, opened 2024-06-13
Add EXCLUDE keyword into CMakeLists

JuanSapriza:
Because in some eXtensions you need a parallel compilation flow (gcc for the main application, and a custom compiler for kernels, for example), we needed a way of excluding files from the compilation flow. 

This is achieved by adding the keyword `exclude` into a directory/file to avoid compilation with gcc. 

Waiting for @LuigiGiuffrida98 to test this on HEEPatia and confirm it works.


JuanSapriza:
Also made a parameter optional in the Makefile (request from @LuigiGiuffrida98 )

LuigiGiuffrida98:
Tested and working up to now, can be merged

PR #523, opened 2024-06-03
SPI Host: HAL and SDK

llorenc-m:
- Fully implemented HAL
- Fully implemented SDK
- SDK example application
- Documentation on SPI Host HAL and SDK

PR #528, opened 2024-06-13
Added DMA SDK and its corresping example

JuanSapriza:
Added the DMA SDK made by (the amazing) @StMiky, also added an example so it can be tested. 
Needed to make some modifications to make it work with the 2D DMA. Please be aware when trying to use it into different repositories. The changes should be transparent to the applications tho.

### DIFF
--- a/sw/applications/example_dma_sdk/main.c
+++ b/sw/applications/example_dma_sdk/main.c
@@ -1,4 +1,4 @@
-// Copyright 2024 EPFL and Politecnico di Torino.
+// Copyright 2024 EPFL
 // Solderpad Hardware License, Version 2.1, see LICENSE.md for details.
 // SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
 //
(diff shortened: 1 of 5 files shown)


sw/applications/example_dma_sdk/main.c
```diff
@@ -0,0 +1,51 @@
+// Copyright 2022 EPFL and Politecnico di Torino.
+// Solderpad Hardware License, Version 2.1, see LICENSE.md for details.
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+//
+// File: dma_sdk.c
+// Author: Juan Sapriza
```
davideschiavone, line 6: maybe can remove Polito from the licence - also, copyright 2024 would be nicer

davideschiavone (changes requested):

JoseCalero:
it is failing in some apps, can you check?
plus, can we follow the same file style as the HAL and SPI SDK (see #523 ) for the .c and .h of the DMA SDK? @JuanSapriza @StMiky

sw/applications/example_dma_sdk/main.c
JuanSapriza: I updated the copyright year. Given that the SDK was written in Polito, I feel it's a shared work. 
Is this relevant in any situation tho? jahaha

PR #543, opened 2024-07-04
fix-malloc

FrancescoPoluzzi:
Fixed brk() and sbrk() syscalls in sw\device\lib\runtime\syscalls.c


JoseCalero:
This is OK. 
I tested it, apparently it fixed #512, i am gonna merge this and then will test again #512 just to double check.

Issue #542, opened 2024-07-03
Fix spi HAL on status access subfields


davideschiavone:

davideschiavone:
#544

PR #544, opened 2024-07-05
Fix SPI byte-addressable problem

JoseCalero:
Related to #542 
The problem is related to the compiler generating 'lb' instructions and the I/O devices not being byte addressable. This is a possible solution, basically it relies on an union to force the compiler having to load the complete word.

- [x] tested with default core
- [x] Need to test on modelsim with cv32e40px

image


JoseCalero:
@davideschiavone it is tested and it works, at least the hellow world with the cv32e40px

PR #551, opened 2024-07-12
Fix #548: use `==` instead of `in` for string comparison in Python/.tpl

### LINKED ISSUE #548: Python: `<string_var> in ("<string>")` should be `<string_var> == "<string>"
This issue affects the following files (that I noticed):
* hw/core-v-mini-mcu/peripheral_subsystem.sv.tpl
* util/mcu_gen.py

For some reason, the Python `in` operator is being used to compare a string variable with a string value, rather than using `==` directly.

These are correct:
```py
s in ("foo", "bar", "baz")
s in ("foo", )
s in ["foo"]
s == "foo"
```
These are not:
```py
s in   "foo"
s in  ("foo")
s in (("foo"))
```
Specifically, `"foo" in ("foolish")` will yield `True` since `"foo"` is a substring of `"foolish"`.

As a result, adding two peripherals with a similar name (e.g. "gpio" and "gpio_improved") can result in one of the peripherals being included twice.

I imagine the use of `in` was justified for comparing a string with multiple possible values, but for single values it should be replaced with `==`.
(Or, if there really is a reason to use `in` instead of `==`, the right-hand side MUST be a list/tuple, e.g. `("foo",)` with a `,` at the end.)

cousteaulecommandant:
The use of `in` for string comparison results in `"name" in "name2"` giving a false positive; the right thing is to use `==`.
In peripheral_subsystem.sv.tpl, this may result in the peripheral being added twice, or being added when it shouldn't.  This was reported in #548, and has been fixed in this PR.
This misuse of `in` also occurs in mcu_gen.py and has been fixed as well.

In addition, this PR also changes the way in which dict keys are handled, from iterating through all the keys to just finding the relevant key, which is more readable (and efficient).

### DIFF
--- a/util/mcu_gen.py
+++ b/util/mcu_gen.py
@@ -220,7 +220,7 @@ class Pad:
     self.pad_mapping = pad_mapping
     self.pad_mux_list = pad_mux_list
 
-    if('low' in pad_active):
+    if pad_active == 'low':
         name_active = 'n'
     else:
         name_active = ''
@@ -626,7 +626,7 @@ def main():
             if isinstance(info, dict):
                 new_info = {}
                 for k, v in info.items():
-                    if k not in ("is_included"):
+                    if k != "is_included":
                         new_info[k] = string2int(v)
                     else:
                         new_info[k] = v
@@ -639,7 +639,7 @@ def main():
         new = {}
         for k,v in peripherals.items():
             if isinstance(v, dict):
-                new[k] = {key:val for key,val in v.items() if key not in ("path")}
+                new[k] = {key:val for key,val in v.items() if key != "path"}
             else:
                 new[k] = v
         return new
@@ -648,10 +648,8 @@ def main():
         len_ep = 0
         for name, info in peripherals.items():
             if isinstance(info, dict):
-                for k, v in info.items():
-                   if k in ("is_included"):
-                    if v in ("yes"):
-                        len_ep += 1
+                if info['is_included'] == "yes":
+                    len_ep += 1
         return len_ep
 
     ao_peripherals = extract_peripherals(discard_path(obj['ao_peripherals']))
@@ -794,7 +792,7 @@ def main():
             pad_mux_list_hjson = []
 
         try:
-            if ('True' in pads[key]['driven_manually']):
+            if pads[key]['driven_manually'] == 'True':
                 pad_driven_manually = True
             else:
                 pad_driven_manually = False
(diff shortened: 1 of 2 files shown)


util/mcu_gen.py
```diff
@@ -219,7 +219,7 @@ def __init__(self, name, cell_name, pad_type, pad_mapping, index, pad_active, pa
     self.pad_mapping = pad_mapping
     self.pad_mux_list = pad_mux_list
 
-    if('low' in pad_active):
+    if pad_active == 'low':
```
davideschiavone, line 222: what if there are commas `,` in the hjson file?

cousteaulecommandant: There shouldn't.  Not in a valid Hjson file at least.
https://hjson.github.io/syntax.html
> **Do not add commas** or comments as they would become part of the string.

Note that I fixed the affected Hjson file in this commit, so that should not be an issue in this particular case.
(But feel free to ditch the second commit of the PR and pull only the first one if you're not sure about that yet.)

Also see issue #552 which I created just now, immediately after submitting this PR, as I realized about that issue while dealing with this one.

davidmallasen:
For some reason this PR didn't trigger the CI. Do you have any ideas why @cousteaulecommandant? I see we also left this one slide and it's related to #638

cousteaulecommandant:
Not really, I don't know why. 
It was dependent on #552 being fixed but if I recall correctly it already included the fix for the relevant files. Maybe CI was manually withheld until #552 was fixed? 

I'll see if I have time to rebase the PR and see if that triggers the CI, although maybe it's wiser to wait for #638 to be merged to avoid/solve any possible merge conflict.

davidmallasen:
Sounds like a plan, let's wait for the other PR to be merged then and rebase to see if it triggers the CI

davideschiavone:
to `dev_branch`

sorry my fault... I made a mess with branches

> Now that #638 is merged -- where do you want me to rebase this? I intended to merge into develop but I think there was a minor mess with the branch (it looks like it was rebased on top of main, and then merged with the remote develop branch rather than force-pushed).
> 
> (I checked the merge conflict and it's a trivial fix; just a couple of commas in pad_cfg.hjson.)

davidmallasen (approved):
Nothing to add from my side. This time the CI triggered and it passed. @davideschiavone is this ok for you? I saw you left a comment at the beginning

PR #555, opened 2024-07-16
adding new FPGA target for apps/hal

davideschiavone:
Most of the legacy SW was targetting PYNQ-Z2 FPGA, however, now we support 3 FPGAs, and the HAL would be broken - 

This PR updates the HAL and APPs with a generic TARGET_IS_FPGA that is defined in each FPGA target


davideschiavone:
@JuanSapriza @JoseCalero  can you pls have a look and test a few applications? (not the power manager ones yet)

JuanSapriza:
@davideschiavone tested all applications (but the power ones) and it worked 👍🏼

PR #560, opened 2024-07-19
Added message inviting people to check they did mcu-gen

JuanSapriza:
Because we are all tired of wasting 15 minutes debugging to see that we didnt mcu-gen... let's have a clear and ironical warning so we do not forget again


JuanSapriza:
@davideschiavone

PR #561, opened 2024-07-20
fix timer SDK

davideschiavone:
just a few fixes, this PR will continue under @FrancescoPoluzzi


davideschiavone:
replaced by #564

PR #575, opened 2024-08-09
make linker folder a parameter

davideschiavone:


Makefile
```diff
@@ -33,6 +33,8 @@ endif
 # Project options are based on the app to be build (default - hello_world)
 PROJECT  ?= hello_world
 
+LINK_FOLDER ?= sw/linker
```
JoseCalero, line 36: should be `LINK_FOLDER ?= $(mkfile_path)/sw/linker`

davideschiavone: yes

sw/device/lib/crt/vectors.S
```diff
@@ -126,39 +126,32 @@ new vector table (which is at mtvec) */
...
 .section .text.vecs
 /* exception handling */
+.globl __no_irq_handler
 __no_irq_handler:
 	la a0, no_exception_handler_msg
 	jal ra, puts
 	j __no_irq_handler
 
 
-sw_irq_handler:
```
JoseCalero, line 135: we could keep the sw handler, in the future could be employed or even for people debugging

PR #584, opened 2024-09-12
Fix clk gating issue (patch)

JoseCalero:
// In Zynq7000, just bypass the clock gating because there are not enough BUFGs that can be 
  // cascaded with the BUFG of the MMCM.
  // In the Zynq UltraScale+, it can be implemented as BUFGCE without trouble, since there
  // are > 500 BUFGCEs and the rules for cascading are more relaxed.
  // NOTE: This **cannot** be substituted by a latch+and


JoseCalero:
@LuigiGiuffrida98 @davideschiavone this should be OK :)

JoseCalero:
@davideschiavone @LuigiGiuffrida98 this was verified in the FPGA, we can merge it

LuigiGiuffrida98:
Fine for me.

PR #586, opened 2024-09-19
Add functionalities to flash bsp

LuigiGiuffrida98:
This PR includes many features and fixes (still adding):

- When DMA is used in polling, as in the FLASH bsp, the DMA hal disables all the interrupts (bit 11 of MIE CSR) without re-enabling them, the fix prevent this to happen disabling only the DMA interrupts (bit 19 of the the MIE CSR). The DMA interrupts are still to be re-enabled at application level. @JoseCalero @JuanSapriza 
- ...

### DIFF
--- a/sw/device/lib/drivers/dma/dma.c
+++ b/sw/device/lib/drivers/dma/dma.c
@@ -60,7 +60,7 @@ extern "C"
 /**
  * Returns the mask to enable/disable DMA interrupts.
  */
-#define DMA_CSR_REG_MIE_MASK (( 1 << 19 ) | (1 << 11 ) ) // @ToDo Add definitions for this 19 and 11
+#define DMA_CSR_REG_MIE_MASK (( 1 << 19 )) // 19 is DMA fast interrupt bit in MIE CSR
 
 /**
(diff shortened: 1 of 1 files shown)


davideschiavone:
@alessionaclerio22 , do not forget to update the documentation

JuanSapriza:
@alessionaclerio22 ToDo for HEEPidermis

.github/workflows/sim-apps-job/test_apps.py
davidmallasen, line 1: @alessionaclerio22 I missed this before, but this PR should not change how the CI tests the applications. Can you revert the changes to this file and check that the PR still passes the tests?

davidmallasen (changes requested):

.github/workflows/sim-apps-job/test_apps.py
alessionaclerio22: @davidmallasen I fixed the CI, but I had to blacklist the test application I developed to test the dma subaddressing mode as it uses the flash.

davidmallasen (approved):

Issue #585, opened 2024-09-18
missing EXTERNAL CRT0 define when compiling internal HEEP apps


LuigiGiuffrida98:
when compiling X-HEEP applications inside a project that vendorizes x-heep, the external crt0 does not get included

davideschiavone:
in my opinion, it the file exists (external crt0) then it should always be included, regardless whether the APP is internal or not @JoseCalero

JoseCalero:
totally agree with @davideschiavone, i will take a look at the cmake backend

davideschiavone:
thanks @JoseCalero

davideschiavone:
@JoseCalero , any news on this?

JoseCalero:
#592 
Done, CMake monster got another head more, it is starting to look like an Hydra

davideschiavone:
2c6d7d8f581020d4961bd08dbd3956c90b676085

PR #592, opened 2024-10-04
missing EXTERNAL CRT0 define 

JoseCalero:
i have been checking this, so when compiling external applications, the external_crt0.S file gets included. Of course, when compiling internal (not external) applications, these are X-HEEP applications, which do not rely on external ctr0. Now, i agree on the fact that we might want to have an external_ctr0.S even for internal applications. This PR fixes that. Moreover, some fixes for verbosity and minor things are also included.

⚠️ Warning! If somebody gets crazy, two external crt files can be added, one externally and other internally (into x-heep). ALWAYS, the one preferred is the outer one.

#585 

This PR has been tested with HEEPatia, HEEPsilon, and bare XHEEP


JoseCalero:
@LuigiGiuffrida98

davideschiavone (approved):

PR #598, opened 2024-11-11
add SREC hex dump

davidmallasen:
Adds the creation of a SREC hex dump to the app flow when the linker is `on_chip`. This can be used to load the binary with JTAG.

P.S. I used this PR to fix a couple of typos


davideschiavone:
@JuanSapriza can you pls have a quick look?

PR #616, opened 2025-01-09
Simulation: fix timestamps on waveform.vcd

cousteaulecommandant:
* Fix timestamps on generated waveform.vcd file so that clock period is 5 ns (100 MHz) rather than 2 ps (500 GHz).
* Add MAX_SIM_TIME parameter to Makefile.
* Add unit suffixes to +max_sim_time (ms, us, ns...).
________

The waveform.vcd file tb_top.cpp generates does not dump the correct timestamps, and just increases the simulation timer by 1 (ps) on every clock edge, resulting in a clock period of 2 ps (500 GHz), instead of 10 ns (100 MHz).  This commit fixes that so that the resulting waveform file has the correct timing.
Also, runCycles(n) was running for n clock edges (i.e. n/2 cycles) rather than n cycles.  Fixed.

Additionally, modify the Makefile so that you can pass a `+max_sim_time=` parameter to the Verilator simulation by setting the variable `MAX_SIM_TIME=` to the desired value.  For example, `make run-app-verilator MAX_SIM_TIME=75000` will stop the simulation if the program hasn't finished after 75000 clock cycles.  This is useful in case our RISC-V application freezes at some point but we don't want to kill the simulation and lose the generated waveform file.
The behavior of this parameter has also been modified so that the simulation can finish earlier if the program finishes before MAX_SIM_TIME has elapsed (this is, on successful program termination, the simulation will finish immediately, rather than waiting for MAX_SIM_TIME to complete).
On premature program termination, the simulation will terminate with an exit status of 2 (rather than 0 or 1).

Finally, add the possibility of adding suffixes to MAX_SIM_TIME to specify time rather than clock cycles:  s, m(s), u(s), n(s), p(s).
With no suffix, use clock cycles.
This way, if you specify `MAX_SIM_TIME=750us` the program will run for 750 microseconds (75000 clock cycles at 100 MHz).
If no unit is specified, assume clock cycles.
Note that only the first letter is considered, so it doesn't matter if you write `1m` or `1ms` or `1meow`.  As a side note, I didn't introduce any kind of error handling, so if you use an invalid letter/character after the integer value (e.g. `1x` or `1N` or `1.0`), no warning/error will be thrown, and the unit will be assumed to be **picoseconds**.  This behavior could (and perhaps should) be modified by adding an `else` at the end.


davideschiavone:
thanks a lot @cousteaulecommandant , looks amazing, can u pls add it to the documentation where running things with verilator is described? so that people know what to pass as command line arguments

tb/tb_top.cpp
```diff
@@ -106,22 +106,29 @@ int main (int argc, char * argv[])
...
 
   if(run_all==false) {
-    runCycles(max_sim_time, dut, m_trace);
+    while(dut->exit_valid_o!=1 && sim_time<max_sim_time) {
+      runCycles(100, dut, m_trace);
+    }
   } else {
     while(dut->exit_valid_o!=1) {
-      runCycles(500, dut, m_trace);
+      runCycles(100, dut, m_trace);
```
davideschiavone, line 114: why 100 instead of 500? any reason?

davideschiavone (changes requested):

tb/tb_top.cpp
cousteaulecommandant: When I fixed the issue with runCycles() only running half as many cycles as required, runCycles(500) would run twice longer (500 cycles instead of 250), so I thought it may make sense to reduce that.
The reason I reduced it to 100 instead of 250 was because it's a good round value and that way the simulation will always run in multiples of 100.  Note that I have also modified the times the initial reset generation uses so that instead of 60 cycles it takes exactly 100 cycles (20 off + 40 on + 40 off) before the main program starts running.  So the entire thing runs in blocks of 100 cycles (one 100-cycle block for the initial reset, and several 100-cycle blocks until program termination or max_sim_time is reached).
I thought this would make it easier for measuring time in the waveform (the program starts at exactly 1 µs simulation time, and is incremented in 1 µs intervals).

Plus, this way the simulation runs "one microsecond at a time", which I think is a nice round number.  (In other words, the value of max_sim_time is interpreted with a precision of 1 µs.)

cousteaulecommandant:
> thanks a lot @cousteaulecommandant , looks amazing, can u pls add it to the documentation where running things with verilator is described? so that people know what to pass as command line arguments

Is there any place in the documentation that mentions `make run-app-verilator`?  I just realized that the documentation just says to call ./Vtestharness directly.  I can add the `+max_sim_time=` parameter there, but it would be nice to document the `run-app-verilator` target.

davideschiavone:
> > thanks a lot @cousteaulecommandant , looks amazing, can u pls add it to the documentation where running things with verilator is described? so that people know what to pass as command line arguments
> 
> Is there any place in the documentation that mentions `make run-app-verilator`? I just realized that the documentation just says to call ./Vtestharness directly. I can add the `+max_sim_time=` parameter there, but it would be nice to document the `run-app-verilator` target.

for now you can extend with your parameters this section, https://x-heep.readthedocs.io/en/latest/How_to/Simulate.html#compiling-for-verilator

cousteaulecommandant:
OK done.  I have added a section describing simulation plusargs to Simulation.md, since in principle `+max_sim_time` applies to all simulations, not only Verilator.  Right now I have only documented `+firmware` (which was already mentioned in the document) and `+max_sim_time`; it might be a good idea to add documentation for `+openOCD` and `+boot_sel` as well in a future.
I also mention that you can use `MAX_SIM_TIME` if using the Makefile, but only for the Verilator simulation.

I have also modified the parsing of time units so that if you enter an invalid character you get an error and the simulation terminates, rather than assuming picoseconds as I was originally doing.

davideschiavone:
> OK done. I have added a section describing simulation plusargs to Simulation.md, since in principle `+max_sim_time` applies to all simulations, not only Verilator. Right now I have only documented `+firmware` (which was already mentioned in the document) and `+max_sim_time`; it might be a good idea to add documentation for `+openOCD` and `+boot_sel` as well in a future. I also mention that you can use `MAX_SIM_TIME` if using the Makefile, but only for the Verilator simulation.
> 
> I have also modified the parsing of time units so that if you enter an invalid character you get an error and the simulation terminates, rather than assuming picoseconds as I was originally doing.

LGTM thanks a lot - as soon as the CI ends I will merge it

PR #618, opened 2025-01-10
Makefile: remove superfluous `@` after `\`

cousteaulecommandant:
Remove superfluous `@` in Makefile recipe lines that are split into multiple lines using `\`:  only the first line should start with `@` and it will apply to all the lines in the multi-line "recipe line"; if any other line in the "recipe line" starts with `@` it will be considered as part of the command, resulting in errors such as `/bin/sh: 3: @echo: not found`.

https://www.gnu.org/software/make/manual/html_node/Splitting-Recipe-Lines.html

As a side note, it might be a good idea to indent all the "continuation lines" in Makefile recipes with one extra tab to make it visually clear that it's all part of a single block, and that an `@` on the first line affects the entire block.  (Vim seems to add that indentation by default so it's probably not a bad idea to follow that convention.)


davidmallasen:
Good catch @cousteaulecommandant. I think for now we can keep the lines without extra indentation if it's not a widely used formatting and it doesn't add any functionality.

PR #619, opened 2025-01-16
Expose cve2 parameters

davidmallasen:
@davideschiavone I left the `cpu_subsystem.sv` without including in the `.gitignore` same as we have the `core_v_mini_mcu.sv`. If you have a different opinion on this let me know. I tested it with and without adding the parameters to the `mcu_cfg.hjson`.


util/mcu_gen.py
```diff
@@ -1073,6 +1083,8 @@ def len_extracted_peripherals(peripherals):
     kwargs = {
         "xheep"                            : xheep,
         "cpu_type"                         : cpu_type,
+        "cve2_rv32e"                       : cve2_rv32e,
```
davideschiavone, line 1086: Can we also add them in the hjson @davidmallasen ?

davideschiavone (changes requested):

davideschiavone:
We have a different way now to set the FPU in cv32e40p*, and cve2 parameters,  I think we should find a way to align them @davidmallasen

davidmallasen:
I want to change that with the new mcu-gen system in the CPU abstraction. I'd rather not add it now and wait to have the definitive version in place to not "break" things twice.

util/mcu_gen.py
davidmallasen: They are not needed and the default behavior is the same as right now, so I think they would add confusion to people that want to change the CPU parameter there. They are optional and documented in the new CPU configuration documentation. This should get better with the new mcu-gen

PR #633, opened 2025-01-31
Verification of SPI slave peripheral

JuanSapriza:
In this PR i try to verify the following

## SPI slave example works properly
The base example was working but would crash if: 
* You added a printf somewhere in the code
* You used 8-bit arrays
I organized the process a bit, removing the complex tests and leaving a simple read/write test. Here you can use any arbitrary length of array in bytes, as well as arrays in half-word or words (as long as you pass the correct size in bytes). 
Because the SPI salve only transmits full words, the received last word is masked and stored in memory protecting the following information (reading, masking, OR-ing with the new data). 

:green_square:  Simulation in verilator works for read and write
:red_square:  Simulation in Questasim
:black_square_button: Changing the number of dummy cycles. Now they had to be fixed to 32 (the default). Changing it crashes the transmission. 
:green_square: Non-buffered transmission. Now we can request single words/bytes at a time. For bytes, full words are transmitted and the masking/shifting is done in SW. 

## FPGA implementation

:green_square: Modified the fpga wrapper and constraints file to expose the pins. 
:green_square: Test example on the FPGA

## Test with external device
:black_square_button: Use an Arduino to read from X-HEEP's memory


JuanSapriza:
@davideschiavone 
I checked this with verilator and on the FPGA and both work fine, yet on Questasim something is not working. My initial suspicion is that it is writing on the wrong address in memory. 
I am using the on-chip linker, because in the testbench (i guess) the pins are set to use on-chip, and therefore other linking would (and indeed does) crash the program. 

Anything I should know about the difference between simulating on Verilator and Questasim before I spend a few hours debugging this?

JuanSapriza:
image

davideschiavone:
I will accept this into the spi branch and then I fix that branch for questasim, with both linker scripts. There should be no reason why it doesn't work but I will check

PR #638, opened 2025-02-05
Hjson: quoteless strings shall not include a comma (#552)

davidmallasen:
Replaces #553


davidmallasen:
@cousteaulecommandant Can you double-check that the changes here are correct? With that, I'll merge when the CI passes

cousteaulecommandant:
I can't test it right now (my development computer is in maintenance) but it looks good to me.

davidmallasen:
No problem, we can wait until you can double-check since this is not blocking (except for your other PR)

.github/workflows/simulate.yml
```diff
@@ -16,8 +16,8 @@ jobs:
           source /root/.bashrc
           conda activate core-v-mini-mcu
           make clean-all
-          sed 's/is_included: "no",/is_included: "yes",/' -i mcu_cfg.hjson
-          sed 's/num_channels:   0x1,/num_channels:   0x4,/' -i mcu_cfg.hjson
-          sed 's/num_channels_per_master_port: 0x1,/num_channels_per_master_port: 0x4,/' -i mcu_cfg.hjson
+          sed 's/is_included: "no"/is_included: "yes"/' -i mcu_cfg.hjson
+          sed 's/num_channels:   0x1$/num_channels:   0x4/' -i mcu_cfg.hjson
```
JuanSapriza, line 20: @davidmallasen  is this `$` here intentional?

cousteaulecommandant: I think that was mine.  I probably added the "end of line" `$` mark to ensure that the line ended in 0x1 and wasn't 0x10 or similar (in which case the match would have failed and no replacement would've been done, but it would fail too in the old form).  I suppose that's never going to happen (it should NOT happen) so you may remove it if you want.

cousteaulecommandant: (just to clarify, the `sed` command still works with the `$` and will match the end of line; it wasn't a typo)

davidmallasen: Yeah, I checked and tested that it worked. Since it wasn't part of the initial behavior and it should work in any case, I removed it to avoid any further confusion. Thanks for the clarifications @cousteaulecommandant!

davidmallasen:
Thank you @cousteaulecommandant! As soon as the CI passes (I had to rebase with some changes on the hjson of the DMA), I will merge

PR #649, opened 2025-02-21
Modified compiler flags to support THEAD Matrix Extensions

Camma2001:
Verified with clang compiler.
To be checked with GCC, but it should not be affected.


sw/CMakeLists.txt
```diff
@@ -368,7 +368,11 @@ else()
     -DportasmHANDLE_INTERRUPT=vSystemIrqHandler\
   ")
 endif()
-set(CMAKE_C_FLAGS ${COMPILER_LINKER_FLAGS})
+
+if (${CMAKE_SYSTEM_PROCESSOR} MATCHES "rv32imc_xtheadmatrix0p1")
+  set(CMAKE_C_FLAGS "${COMPILER_LINKER_FLAGS} -menable-experimental-extensions")
```
davidmallasen, line 373: You only set the `CMAKE_C_FLAGS ` in the THEAD case? Please test this with the current configurations too. I believe this breaks the normal flow

sw/cmake/riscv.cmake
```diff
@@ -101,6 +101,9 @@ endif()
 # Set the CMAKE C flags (which should also be used by the assembler!
 set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g" )
 set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=${CMAKE_SYSTEM_PROCESSOR}" )
+if (${CMAKE_SYSTEM_PROCESSOR} MATCHES "rv32imc_xtheadmatrix0p1")
```
davidmallasen, line 104: We should make this generic so that it only looks for the `xtheadmatrix` substring instead of the exact march

davidmallasen (changes requested):

davideschiavone:
fixed in 643b369efc8726f6c18b369da39f7e0abe444a4f

PR #664, opened 2025-03-04
Fixed DMA bug & improved im2col SPC

TommiTerza:
- Fixed a bug in _dma_pdding_fsm.sv_, which caused the DMA, in certain situations, to wrongly pad, causing the FIFOs to not be correctly empties and corrupting the successive DMA transactions
- Improved the computation of the number of zeros for each DMA transaction, reducing the complexity of the im2col accelerator


davideschiavone:
@TommiTerza - any news on the FPGA bug?

Issue #550, opened 2024-07-12
spi_sdk APP is not working


davideschiavone:

davideschiavone:
@JoseCalero , any news on this?

JoseCalero:
not checked yet, i will check it this friday

PR #667, opened 2025-03-10
SPI slave work on Questasim

JuanSapriza:


JuanSapriza:
Tried simulating all with questasim

```
 FAIL: 1 APPS COULD NOT BE BUILT
 FAIL: 7 APPS FAILED THE SIMULATION
(Timeout)       coremark
(questasim)     example_dma_subaddressing
(Timeout)       example_freertos_blinky
(gcc,on_chip)           example_pdm2pcm
(Timeout)       example_sdk_spi_flash
(questasim)     example_spidma_powergate
(Timeout)       example_spi_read
(Timeout)       minver
```

davideschiavone:
> Tried simulating all with questasim
> 
> ```
>  FAIL: 1 APPS COULD NOT BE BUILT
>  FAIL: 7 APPS FAILED THE SIMULATION
> (Timeout)       coremark
> (questasim)     example_dma_subaddressing
> (Timeout)       example_freertos_blinky
> (gcc,on_chip)           example_pdm2pcm
> (Timeout)       example_sdk_spi_flash
> (questasim)     example_spidma_powergate
> (Timeout)       example_spi_read
> (Timeout)       minver
> ```

probably example_spi_read uses the the host spi to talk with the FLASH - and now it is not connected to it anymore -

JuanSapriza:
@davideschiavone 

currently the logic is: 
* if in simulation, use the SPI host to get tdata -no longer available -  
* if i FPGA use the SPI flash to communicate with the flash on the programmer. 

If i simply remove the option of the SPI host then none of these applications will be simulate-able. Why cant we use the spi flash in simulation? The flash is in the testharness connected to the spi flash for flash_load/exec simulation. It would not work on LINKER=flas_load/exec, but couldnt it work with LINKER=on_chip?

JuanSapriza:
yeah yeah, but what i mean is that, until the use of SPI_FLASH was restricted to FPGA 
```c
#if defined(TARGET_PYNQ_Z2) || defined(TARGET_ZCU104) || defined(TARGET_NEXYS_A7_100T)
    #define USE_SPI_FLASH
#endif
```

I was wondering if there was a reason not to use SPI_FLASH in simulation

davideschiavone:
> yeah yeah, but what i mean is that, until the use of SPI_FLASH was restricted to FPGA
> 
> ```c
> #if defined(TARGET_PYNQ_Z2) || defined(TARGET_ZCU104) || defined(TARGET_NEXYS_A7_100T)
>     #define USE_SPI_FLASH
> #endif
> ```
> 
> I was wondering if there was a reason not to use SPI_FLASH in simulation

the reason was that we took the chance to test the SPI_HOST that way, otherwise we risked that we would never test it

JuanSapriza:
Fantastic, then I will move all apps to SPI_FLASH and test them - fingers crossed and they will work :crossed_fingers:

tb/tb_top.cpp
```diff
@@ -115,15 +115,19 @@ int main (int argc, char * argv[])
     }
   }
 
-  if(dut->exit_valid_o==1) {
+  std::cout<<"Simulation finished after "<<(sim_time/CLK_PERIOD_ps)<<" clock cycles"<<std::endl;
```
davideschiavone, line 118: why did you move this year?

JuanSapriza: because otherwise the `make app-simulate-all` does not work hahaahahah (which tells you how much people have been testing their PRs hahaha

The test-all.sh checks the last character to be 0, if we print something else after that then it does not work.

JuanSapriza:
To have the applications using  SPI_FLASH to work on verilator i needed to add the model (include it in the testharness), but now the simulation model takes an eternity to build (it would take 2/3 minutes before, now it has been at least 10min and it hasnt finished). Even if it works fine, this pretty undesirable

tb/tb_top.cpp
davideschiavone: we should not rely on the UART output but on the return value... let's revert this change and discuss this with @davidmallasen in the other PR, for this PR we should revert this part pls

JuanSapriza:
I am guessing that is the reason the spiflash memory was not used with verilator. 

I am now running simulation of all apps with questasim. 

```
 ================================================================================
 Successfully built demo_spi_master_slave using gcc for on_chip
 ================================================================================
make[1]: *** [run] Error 2
 ================================================================================
 Successfully simulated demo_spi_master_slave using questasim
 ================================================================================
 
  ================================================================================
 Successfully built example_sdk_spi_flash using gcc for on_chip
 ================================================================================
make[1]: *** [run] Error 2
 ================================================================================
 Successfully simulated example_sdk_spi_flash using questasim
 ================================================================================
 
  ================================================================================
 Successfully built example_spi_write using gcc for on_chip
 ================================================================================
  Timeout!
 ================================================================================
 Timeout simulating example_spi_write using questasim
 ================================================================================
 
 ```

tb/tb_top.cpp
JuanSapriza: this is not a UART output, it's the return value `return 0` that the makefile parses and pushes oupward to be captured by the script. 
In the script from the other PR they are doing something similar but more sophisiticated (i.e. they have a regex to capture this value, instead of depending on catching the last character, so they dont care about the order)

davidmallasen: We do use the return value for the tests and not the stdout prints

JuanSapriza:
✅ demo_spi_master_slave
✅ example_spi_write
❌ example_sdk_spi_flash

The last example standing has a problem with trying to read from an address with Xs. I havent been able to pinpoint from where exactly this is coming yet.

davideschiavone:
> ✅ demo_spi_master_slave ✅ example_spi_write ❌ example_sdk_spi_flash
> 
> The last example standing has a problem with trying to read from an address with Xs. I havent been able to pinpoint from where exactly this is coming yet.

That was already broker, due to passing pointers instead of variables or something similar https://github.com/esl-epfl/x-heep/issues/550

So if that one is the only broken one, i merge, then we have to fix synthesis in another PR

JuanSapriza:
Alright! Then let's merge!

PR #668, opened 2025-03-11
Update app testing script

davidmallasen:
Updates the script that tests the execution of all of the applications. This is used in the CI.


davideschiavone:
@davidmallasen please add also targets to test the APPs (at least compilation) with the OpenHW Group compiler - @LuigiGiuffrida98 or @StMiky , would it be possible to update the docker with the OpenHW Group compiler?

LuigiGiuffrida98:
Yes, I already have it, let me update the action to point to the updated docker

davidmallasen:
For that we need to add it to the docker. I'm separating this part of updating the CI so we can merge it ASAP, and we can add the gcc and docker stuff in the other PR. I would suggest to do it in smaller increments so that it doesn't end up being huge PRs that take a long time to be finished and merged

davidmallasen:
There you go @JuanSapriza. I removed your `.sh` script as I think it gets replaced by the python script. Let me know if there was a use for it that is not covered by the python so we can add it.

PR #685, opened 2025-03-27
[Profiling] Integrate `rv-profile` tool into X-HEEP

Vincenzo-Petrolo:
Hi, 
This PR has the goal to bring [rv-profile ](https://github.com/vlsi-lab/rv_profile) into X-HEEP. 

## Graphic example of a FlameGraph
This is an example of what is generated after running the FFT example in X-HEEP.
FlameGraph

## Features
- All the CPUs currently supported by X-HEEP are also supported by [rv-profile ](https://github.com/vlsi-lab/rv_profile)
- Generation of interactive FlameGraphs. Use the example [here](https://vincenzo-petrolo.github.io/flamegraph_example/flamegraph.svg)

## TODO
Among the things missing I see (feel free to edit and add more):
- [x] Adding a page to the doc


davidmallasen:
Hello @Vincenzo-Petrolo. Could you move `profile` to `util/profile`? I prefer not having a new top-level directory. Then, you can also keep the script in the same profile directory to avoid having it scattered in two different places

Vincenzo-Petrolo:
Hi @davidmallasen, I did it. 

Please let me know If anything else needs to be addressed.

davidmallasen:
Thanks @Vincenzo-Petrolo. Then I think we're just missing the docs page documenting it. I would suggest adding a profiling page under how to. If you need more guidance let me know

Vincenzo-Petrolo:
Hi @davidmallasen, I added the page "Profiling" under "How To" 👍🏻.

davidmallasen:
Hello @Vincenzo-Petrolo. I tried running this with the information from the docs and I run into this:
~~~
usage: rv_profile [-h] --elf ELF --fst FST --cfg CFG [--out OUT]
rv_profile: error: argument --fst: expected one argument
~~~
So I believe something is missing. For example, I didn't know if I had to add a specific compile/execution flag or what simulator I had to use. I tried compiling and running the `example_matadd` with verilator.

davideschiavone:
@Vincenzo-Petrolo - what is missing to merge this PR?

davideschiavone:
> Hi @davideschiavone, yes I know about this. Unfortunately this is due to the wal-lang package which I cannot change.
> 
> However if you install the rv-profile using `make conda` it will install a custom fork from my git repository that works with python3.8. Hence, if you install rv-profile through pip it won't work. Try to do that using `make conda`.

I still have the same issue after a fresh `make venv`

Issue #688, opened 2025-04-03
[docs] SPI and SPI2 are both in the peripheral subsystem, not in AO


cousteaulecommandant:
The documentation and diagram claims that the "SPI" block is placed in the Always-on peripheral subsystem along with "SPI Flash", whereas "SPI 2" is in the Peripheral subsystem.
However, it seems that both SPI and SPI 2 are in the Peripheral subsystem with SPI Flash being the only SPI peripheral in the AO region (with two multiplexed control interfaces: spi_memio which provides memory-mapped DMA access, and spi_flash which doesn't), as shown in mcu_cfg.hjson.

Affected files:
* docs/source/index.rst
* docs/source/images/x_heep.svg
* docs/source/images/xheep_diagram.png

As a side note, while you're modifying index.rst, I would also suggest to rearrange the sections so that "Always-on peripheral subsystem domain" is introduced first, and then "Peripheral subsystem domain" after it.

Here are some suggestions for the layout of the modified diagram (depending on whether or not you decide to add the SPI MEMIO block as well, although I understand that "SPI FLASH" here refers to the whole SPI subsystem):
```
   AO PERI SUBSYS                              PERI SUBSYS
--+-------+-------+-------+-------+-------+   +-------+-------+-------+
--|   S   |   S   |   S   | S M M |   S   |   |   S   |   S   |   S   |--
  |  SOC  | POWER |  BOOT |       |       |   |       |       |       |--
--|  CTRL |  MGR  |  ROM  |  DMA  | TIMER |   |TIMER 2| PLIC  |  I2C  |
--+-------+-------+-------+-------+-------+   +-------+-------+-------+
  |   S   |   S   |   S   |S_FL  S|   S   |   |   S   |   S   |   S   |
  |F.INTR |       |       |  SPI  |  SPI  |   |       |       |       |
  | CTRL  | UART  |GPIO AO| MEMIO | FLASH |   |  SPI  | SPI 2 | GPIO  |
  +-------+-------+-------+---\_______/---+   +-------+-------+-------+
             | |      |         | | |           | | |   | | |     |
```
```
   AO PERI SUBSYS                      PERI SUBSYS
  +-------+-------+-------+-------+   +-------+-------+
--|   S   |   S   | S M M |   S   |   |   S   |   S   |--
--|       |  BOOT |       |F.INTR |   |       |       |--
  |       |  ROM  |       | CTRL  |   | PLIC  |  I2C  |
[...]

PR #689, opened 2025-04-09
Update diagram in README

JuanSapriza:
The diagram in the readme had a small mistake (reported in #688) + was not helping understand how the different buses work together. 
I added this editable diagram were we can update any time something new is added or something is modified. 

Comments are welcomed :)


davideschiavone:
lgtm, can I merge @JuanSapriza ?

JuanSapriza:
be my guest :)

cousteaulecommandant:
This new diagram is much more detailed indeed, but it seems a bit confusing.  Here are some suggestions:

* Arrows are a bit messy.  They appear to be mixing "bus interfaces" with "individual signals" (mostly interrupts), which I think should be clearly differentiated (consider using colors or dotted lines?) since they have an entirely different nature (a "32-bit interface" is not the same as 32 individual bits - it's bidirectional and includes address and control signals as well).
  In fact, I would entirely remove the "individual signals" since I think they contribute very little and make the diagram more cluttered.  (The version with all the arrows is still useful as a more complete version, but perhaps the one shown in the main README should be simplified and show only the bus interfaces.)

* I liked it more when all different subsystems were highlighted.  I see that this version has been kept as "picture 2/2"; maybe it should be the default.  (Once the interrupts and power manager signals have been removed, the diagram will be cleaner.)

* The background of "custom CHEEP" should be white/transparent to make it clearer that it is not part of the X-HEEP project itself (and maybe make the components semi-transparent or light-colored)

Here's a possible arrow format to differentiate all types of connection:


And an example of what the diagram looks like with only the interface arrows (much cleaner in my opinion; a version with all the arrows could still be provided as a more detailed diagram):

PR #686, opened 2025-03-28
Add digital Level Crossing block as IP example

alessionaclerio22:
### PR Goal
This PR integrates a digital Level Crossing (dLC) block to the ip examples in x-heep. The dLC is meant to be tightly-coupled to the DMA, working in streaming exploiting the DMA HW Fifo mode.

### What's new

1. dLC IP in `hw/ip_examples/dlc`
2. dLC test application in `sw/applications/example_dlc`

### TODOs

- [x] Modify golden dLC results in the .h file in `sw/applications/example_dlc` because the structure of the dLC response packet has been changed to be more flexible.
- [x] Add the dLC test application to the CI.

PR #699, opened 2025-04-18
Re-structured DMA to allow its customization

TommiTerza:
The main goal of this PR is to restructure the DMA system to allow for a fine-grained customization. It is now possible to choose each DMA feature, such as Address Mode, Subaddressing Mode, Zero-Padding and HW FIFO Mode, to suit a wider performance-area design space.
Furthermore, this restructuring improved the efficency of the lastest features, like HW FIFO and Subaddressing Mode, as well as simplifying the existing sign-extension logic and improving the overall robustness of the system.
The im2col SPC has been improved too, with an optimized logic that reduces even further its lightweight area footprint.
Last but not least, the DMA system has been modified to be fully OBI-compliant.

To do:

- [ ] Wait for PR #686 to be merged to rebase and add minor modifications
- [ ] Update the documentation


hw/core-v-mini-mcu/include/fifo_pkg.sv
```diff
@@ -5,21 +5,22 @@
 // File: hw_fifo_pkg.sv
 // Author: Alessio Naclerio
 // Date: 17/02/2025
-// Description: Package for HW FIFO MODE dma interface.
+// Description: Package for FIFOs in the DMA subsystem
 
-package hw_fifo_pkg;
+package fifo_pkg;
```
davideschiavone, line 10: Fifo is too generico, let's call them hw_dma_fifo

davideschiavone (changes requested):

hw/core-v-mini-mcu/include/fifo_pkg.sv
TommiTerza: This interface is used for the internal FIFOs too, that’s why it’s so generic. We can call it dma_fifo, but it could be applied to any FIFO interface, so we might want to let it as generic as it is

davideschiavone:
hi @TommiTerza , I think you need to merge the main branch into yours and this is now unmergiable

hw/ip/dma/data/dma.hjson
JuanSapriza, line 1: @TommiTerza shouldnt the registers that are not used (given a DMA channel configuration) be removed? or we keep a general register topology for all DMA channels and we pay the cost of some "useless" registers?

TommiTerza: Good question! The hw/ip/dma/data/dma.hjson is the full structure of the DMA registers. Whenever **mcu_gen** is called, the _dma.hjson_ is processed into a _dma_gen.hjson_ (untracked by git): for every mode disabled, registers are removed. Then, _regtool_ is called to generate both hw and sw. 
In this way, the memory map of the registers is updated transparently to the user, and sw keeps on working like always.

davidmallasen: To follow the same structure than everywhere else, could we call this `dma.hjson.tpl` and `dma.hjson`?

davidmallasen: You can do it similarly to the power manager: https://github.com/esl-epfl/x-heep/tree/main/hw/ip/power_manager/data and reuse `mcu-gen` as done there

TommiTerza: Yes, I can make it so, no probelm

.Xil/Vivado-87972-Tommi/hdfFromDcp/xilinx_core_v_mini_mcu_wrapper.hwdef
JuanSapriza, line 1: Please remove the .Xil folder

JuanSapriza (changes requested):

davidmallasen:
Is the `docs/source/images/dma_structure.png` still needed? If not, let's delete it

sw/applications/example_dma_2d/main.c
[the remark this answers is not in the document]
JuanSapriza: false alarm... you cannot do that because it's a macro. 

What i did was remove it and define on-demand whenever it was required. 

I also guarded the test_address_mode by some `#if`
```c
#if DMA_ADDR_MODE
#define TEST_ADDRESS_MODE
//#define TEST_ADDRESS_MODE_EXTERNAL_DEVICE
#endif
```

hw/core-v-mini-mcu/core_v_mini_mcu.sv.tpl
JuanSapriza, line 1: @alessionaclerio22 @TommiTerza @davideschiavone @davidmallasen 
From what i remembered we were going to connect the window interrupt of the DMA to the fast intr controller. 
With this in mind i completely removed the plic from HEEPidermis. 

If it is not much work, i would ask you (@TommiTerza) to go ahead and connect the window interrupt to the fic :pray:

JuanSapriza (changes requested):
If i apply the changes i mention and add the plic back in HEEPidermis, then the example_dma works. 
I will be testing more complex examples, but maybe it's worth to fix these things, as I will like to test the window mode without the PLIC :)

JuanSapriza:
# This has been tested on HEEPidermis and works like a charm
Including the window count not resetting in circular mode :+1: 
Very nice work @TommiTerza !!! :partying_face: 

Only change i needed to do was in the dma.core:
```
 - dma_zero_padding ? (rtl/dma_units/dma_processing_unit.sv)
 - dma_addr_mode ? (rtl/dma_units/dma_read_addr_unit.sv)
```

So these blocks are not compiled in _other_ simulators.

JuanSapriza:
Also... for the record... **GREAT DOCUMENTATION!** 
I was going to write to tommaso with a few questions, but went to the docu and it even has a nice index at the beginning with the things you are going there for hhaha nice!!!!!

TommiTerza:
@JuanSapriza thank you so much!
Glad to see that the documentation is as usegul as I hoped!

alessionaclerio22:
I tested the dma enabling all the modes in d-heep-blue (which is heepatia basically). I specifically tested the hw_fifo mode using my cgra and it works perfectly (both verilator and questasim)!! I also checked other dma-related applications and it seems fine!

hw/ip_examples/dlc/rtl/dlc.sv
```diff
@@ -100,6 +102,20 @@ module dlc #(
 
   // ------------------------- FSM
 
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (~rst_ni) begin
+      trans_counter <= reg2hw.trans_size.q;
```
davideschiavone, line 107: @alessionaclerio22 .....

davideschiavone (changes requested):

davideschiavone (approved):

PR #707, opened 2025-05-01
Simplify kwargs

Pacsort17:
Goal : Remove all peripheral related variables in kwargs, every mako template should call xheep object to get these values

Variables removed : 
    ao_peripheral_start_address
    ao_peripheral_size_address
    ao_peripherals
    ao_peripherals_count
    dma_ch_count
    dma_ch_size,
    num_dma_master_ports
    num_dma_xbar_channels_per_master_port
    dma_xbar_masters_array
    peripheral_start_address
    peripheral_size_address
    peripherals
    peripherals_count

Files changes : 
    hw/core-v-mini-mcu/include/core_v_mini_mcu_pkg.sv.tpl
    hw/core-v-mini-mcu/system_bus.sv.tpl
    sw/device/lib/runtime/core_v_mini_mcu.h.tpl
    hw/ip/power_manager/data/power_manager.hjson.tpl
    hw/core-v-mini-mcu/peripheral_subsystem.sv.tpl
    hw/core-v-mini-mcu/core_v_mini_mcu.sv.tpl
    hw/ip/power_manager/rtl/power_manager.sv.tpl

Also part of the PR : 
 - Computing dma xbar array in the dma object (computed each time the corresponding get function is called), not in mcu_gen anymore
 - Extending load_config file with a function that takes a file as the "default" configuration (add all missing peripherals in xheep object). Note : when configuration files will be merged, the call in mcu_gen.py to this function should be moved in load_config.py (currently impossible, since load_config.py doesn't know anything about --cfg_peripherals argument)
 - Adding helper function in PeripheralDomain to check if a peripheral is already in the domain or not
 - Removing almost all logic concerning peripherals in mcu_gen.py
 - Doesn't generate not included peripherals in .sv files anymore (especially hw/core-v-mini-mcu/include/core_v_mini_mcu_pkg.sv)

Note : since the docker has a version of python prior to 3.10, match keyword doesn't exist and is replaced with a big if / elif / else block

### DIFF
--- a/util/x_heep_gen/load_config.py
+++ b/util/x_heep_gen/load_config.py
@@ -2,7 +2,36 @@ import importlib
 from pathlib import PurePath
 from typing import List, Optional, Union
 import hjson
+import os
+import sys
+from jsonref import JsonRef
 
+from .peripherals.base_peripherals import (
+    BasePeripheralDomain,
+    SOC_ctrl,
+    Bootrom,
+    SPI_flash,
+    SPI_memio,
+    DMA,
+    Power_manager,
+    RV_timer_ao,
+    Fast_intr_ctrl,
+    Ext_peripheral,
+    Pad_control,
+    GPIO_ao,
+    UART,
+)
+from .peripherals.user_peripherals import (
+    UserPeripheralDomain,
+    RV_plic,
+    SPI_host,
+    GPIO,
+    I2C,
+    RV_timer,
+    SPI2,
+    PDM2PCM,
+    I2S,
+)
 from .linker_section import LinkerSection
 from .system import BusType, Override, XHeep
 
@@ -177,6 +206,130 @@ def load_linker_config(system: XHeep, config: list):
             end = None
 
         system.add_linker_section(LinkerSection(name, start, end))
+
+
+def load_peripherals_config(system: XHeep, config_path: str):
+    """
+    Reads the whole peripherals configuration.
+
+    :param XHeep system: the system object where the peripherals should be added.
+    :param str config_path: The path to the configuration file.
+    :raise ValueError: If config file does not exist or if peripheral name doesn't match a peripheral class.
+    """
+
+    if not os.path.exists(config_path):
+        raise ValueError(
+            f"Peripherals configuration file {config_path} does not exist."
+        )
+
+    with open(config_path, "r") as file:
+        try:
+            srcfull = file.read()
+            config = hjson.loads(srcfull, use_decimal=True)
+            config = JsonRef.replace_refs(config)
+        except ValueError:
+            raise SystemExit(sys.exc_info()[1])
+
+    for name, fields in config.items():
+        # Base Peripherals
+        if name == "ao_peripherals":
+            base_peripherals = BasePeripheralDomain(
+                int(fields["address"], 16), int(fields["length"], 16)
+            )
+            # iterate over all peripherals and create corresponding objects
+            for peripheral_name, peripheral_config in fields.items():
+                if peripheral_name == "address" or peripheral_name == "length":
+                    continue
+
+                offset = int(peripheral_config["offset"], 16)
+                length = int(peripheral_config["length"], 16)
+                if peripheral_name == "soc_ctrl":
+                    peripheral = SOC_ctrl(offset, length)
+                    peripheral.custom_configuration(peripheral_config["path"])
+                elif peripheral_name == "bootrom":
+                    peripheral = Bootrom(offset, length)
+                elif peripheral_name == "spi_flash":
+                    peripheral = SPI_flash(offset, length)
+                elif peripheral_name == "spi_memio":
+                    peripheral = SPI_memio(offset, length)
+                elif peripheral_name == "dma":
+                    peripheral = DMA(
+                        offset,
+                        length,
+                        int(peripheral_config["ch_length"], 16),
+                        int(peripheral_config["num_channels"], 16),
+                        int(peripheral_config["num_master_ports"], 16),
+                        int(peripheral_config["num_channels_per_master_port"], 16),
+                    )
+                    peripheral.custom_configuration(peripheral_config["path"])
+                elif peripheral_name == "power_manager":
+                    peripheral = Power_manager(offset, length)
+                    peripheral.custom_configuration(peripheral_config["path"])
+                elif peripheral_name == "rv_timer_ao":
+                    peripheral = RV_timer_ao(offset, length)
+                elif peripheral_name == "fast_intr_ctrl":
+                    peripheral = Fast_intr_ctrl(offset, length)
+                    peripheral.custom_configuration(peripheral_config["path"])
+                elif peripheral_name == "ext_peripheral":
+                    peripheral = Ext_peripheral(offset, length)
+                elif peripheral_name == "pad_control":
+                    peripheral = Pad_control(offset, length)
+                elif peripheral_name == "gpio_ao":
+                    peripheral = GPIO_ao(offset, length)
+                elif peripheral_name == "uart":
+                    peripheral = UART(offset, length)
+                    peripheral.custom_configuration(peripheral_config["path"])
+                else:
+                    raise ValueError(f"Peripheral {peripheral_name} does not exist.")
+                # Adding peripheral to domain
+                base_peripherals.add_peripheral(peripheral)
+
+            # All peripherals in configuration filehave been added
+            system.add_peripheral_domain(base_peripherals)
+        # User Peripherals
+        elif name == "peripherals":
+            user_peripherals = UserPeripheralDomain(
+                int(fields["address"], 16), int(fields["length"], 16)
+            )
+            # iterate over all peripherals and create corresponding objects
+            for peripheral_name, peripheral_config in fields.items():
+                if peripheral_name == "address" or peripheral_name == "length":
+                    continue
+
+                offset = int(peripheral_config["offset"], 16)
+                length = int(peripheral_config["length"], 16)
+                # Skip if the peripheral is not included
+                if peripheral_config["is_included"] == "no":
+                    continue
+                if peripheral_name == "rv_plic":
+                    peripheral = RV_plic(offset, length)
+                    peripheral.custom_configuration(peripheral_config["path"])
+                elif peripheral_name == "spi_host":
+                    peripheral = SPI_host(offset, length)
+                    peripheral.custom_configuration(peripheral_config["path"])
+                elif peripheral_name == "gpio":
+                    peripheral = GPIO(offset, length)
+                    peripheral.custom_configuration(peripheral_config["path"])
+                elif peripheral_name == "i2c":
+                    peripheral = I2C(offset, length)
+                    peripheral.custom_configuration(peripheral_config["path"])
+                elif peripheral_name == "rv_timer":
+                    peripheral = RV_timer(offset, length)
+                    peripheral.custom_configuration(peripheral_config["path"])
+                elif peripheral_name == "spi2":
+                    peripheral = SPI2(offset, length)
+                elif peripheral_name == "pdm2pcm":
+                    peripheral = PDM2PCM(offset, length)
+                    peripheral.custom_configuration(peripheral_config["path"])
+                elif peripheral_name == "i2s":
+                    peripheral = I2S(offset, length)
+                    peripheral.custom_configuration(peripheral_config["path"])
+                else:
+                    raise ValueError(f"Peripheral {peripheral_name} does not exist.")
+                # Adding peripheral to domain
+                user_peripherals.add_peripheral(peripheral)
+            # All peripherals in configuration file have been added
+            system.add_peripheral_domain(user_peripherals)
 
 
 def load_cfg_hjson(src: str, override: Optional[Override] = None) -> XHeep:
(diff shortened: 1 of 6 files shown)


davideschiavone:
please wait for https://github.com/esl-epfl/x-heep/pull/699 to merge this one

hw/core-v-mini-mcu/include/core_v_mini_mcu_pkg.sv.tpl
```diff
@@ -42,7 +42,7 @@ package core_v_mini_mcu_pkg;
   localparam logic [31:0] DMA_WRITE_P0_IDX = 4;
   localparam logic [31:0] DMA_ADDR_P0_IDX = 5;
  
-  localparam SYSTEM_XBAR_NMASTER = ${3 + int(num_dma_master_ports)*3};
+  localparam SYSTEM_XBAR_NMASTER = ${3 + int(hex(xheep.get_dma()[0].get_num_master_ports())[2:])*3};
```
davidmallasen, line 45: Why get this as hex? Isn't it similar to what it's done in `hw/core-v-mini-mcu/core_v_mini_mcu.sv.tpl`?

hw/core-v-mini-mcu/include/core_v_mini_mcu_pkg.sv.tpl
```diff
@@ -82,13 +82,13 @@ package core_v_mini_mcu_pkg;
   localparam logic[31:0] DEBUG_END_ADDRESS = DEBUG_START_ADDRESS + DEBUG_SIZE;
   localparam logic[31:0] DEBUG_IDX = 32'd${xheep.ram_numbanks() + 1};
 
-  localparam logic[31:0] AO_PERIPHERAL_START_ADDRESS = 32'h${ao_peripheral_start_address};
-  localparam logic[31:0] AO_PERIPHERAL_SIZE = 32'h${ao_peripheral_size_address};
+  localparam logic[31:0] AO_PERIPHERAL_START_ADDRESS = 32'h${f"{xheep.get_base_peripherals_base_address() & 0xFFFFFFFF:08X}"};
```
davidmallasen, line 85: Lets change this to `...start_address()` instead of `...base_address()` to keep it consistent with the names, and to avoid confusion with the base peripherals

hw/core-v-mini-mcu/include/core_v_mini_mcu_pkg.sv.tpl
```diff
@@ -82,13 +82,13 @@ package core_v_mini_mcu_pkg;
...
+  localparam logic[31:0] AO_PERIPHERAL_SIZE = 32'h${f"{xheep.get_base_peripherals_length() & 0xFFFFFFFF:08X}"};
   localparam logic[31:0] AO_PERIPHERAL_END_ADDRESS = AO_PERIPHERAL_START_ADDRESS + AO_PERIPHERAL_SIZE;
   localparam logic[31:0] AO_PERIPHERAL_IDX = 32'd${xheep.ram_numbanks() + 2};
 
-  localparam logic[31:0] PERIPHERAL_START_ADDRESS = 32'h${peripheral_start_address};
-  localparam logic[31:0] PERIPHERAL_SIZE = 32'h${peripheral_size_address};
+  localparam logic[31:0] PERIPHERAL_START_ADDRESS = 32'h${f"{xheep.get_user_peripherals_base_address() & 0xFFFFFFFF:08X}"};
```
davidmallasen, line 90: To avoid all these small "utility" functions in the `system.py`, I think we should modularize it. The system can have a `get_base_peripheral_domain()`, and `get_user_peripheral_domain()`, then the BasePeripheralDomain can have the get functions that are needed there, etc. Let's discuss this

davidmallasen (changes requested):

davidmallasen:
@Pacsort17 #699 got merged, so feel free to merge/rebase with those changes and adapt any conflicts to the modifications that you're introducing in this PR

davideschiavone:
@davidmallasen pls guys find a better name for this PR so that the commit message is clear :)

hw/core-v-mini-mcu/include/core_v_mini_mcu_pkg.sv.tpl
```diff
@@ -133,72 +139,75 @@ package core_v_mini_mcu_pkg;
...
+  localparam PERIPHERALS = ${len(user_peripheral_domain.get_peripherals())};
 
-% for peripheral, addr in peripherals.items():
-  localparam logic [31:0] ${peripheral.upper()}_START_ADDRESS = PERIPHERAL_START_ADDRESS + 32'h${addr["offset"]};
-  localparam logic [31:0] ${peripheral.upper()}_SIZE = 32'h${addr["length"]};
-  localparam logic [31:0] ${peripheral.upper()}_END_ADDRESS = ${peripheral.upper()}_START_ADDRESS + ${peripheral.upper()}_SIZE;
-  localparam logic [31:0] ${peripheral.upper()}_IDX = 32'd${loop.index};
+% for peripheral in user_peripheral_domain.get_peripherals():
+  localparam logic [31:0] ${peripheral.get_name().upper()}_START_ADDRESS = PERIPHERAL_START_ADDRESS + 32'h${f"{peripheral.get_address() & 0xFFFFFFFF:08X}"};
+  localparam logic [31:0] ${peripheral.get_name().upper()}_SIZE = 32'h${f"{peripheral.get_length() & 0xFFFFFFFF:08X}"};
```
davidmallasen, line 203: No need for the 0xFFFF...

util/x_heep_gen/load_config.py
```diff
@@ -179,6 +208,179 @@ def load_linker_config(system: XHeep, config: list):
...
+                elif peripheral_name == "uart":
+                    peripheral = UART(offset, length)
+                    peripheral.custom_configuration(peripheral_config["path"])
+                else:
+                    raise ValueError(f"Peripheral {peripheral_name} does not exist.")
+                # Adding peripheral to domain
+                if base_peripherals is not None:
+                    base_peripherals.add_peripheral(peripheral)
+                else:
+                    # TODO : add peripheral to system --> add a function in system to add a peripheral, or add a function to get the domain ?
```
davidmallasen, line 315: Double-check this

Pacsort17: I simplified that code, but the check must be there (otherwise I would have None.add_peripheral since the for loop was executed even if there was no base_peripherals) -> now the for loop only executes only if there is a base_peripheral_domain

util/x_heep_gen/peripherals/abstractions.py
```diff
@@ -80,7 +80,7 @@ def custom_configuration(self, config_path: str):
 
         :param str config_path: The path to the hjson file that describes the peripheral. If the path does not exist, a FileNotFoundError will be raised.
         """
-        if not path.exists(config_path):
+        if not path.exists(config_path) and not path.exists(config_path + ".tpl"):
```
davidmallasen, line 83: @Pacsort17, when testing this PR in our repositories, this extra check fails if we run `mcu-gen.py` from both inside x-heep and outside. Since we don't use these paths from outside x-heep and I think this will be fixed with your next PR, can you comment these two lines (the if) and then we re-add them in your next PR?

davidmallasen (changes requested):

hw/core-v-mini-mcu/peripheral_subsystem.sv.tpl
```diff
@@ -289,7 +293,7 @@ module peripheral_subsystem
       .reg_rsp_o(peripheral_slv_rsp[core_v_mini_mcu_pkg::RV_PLIC_IDX])
   );
 
-% if 'rv_plic' in peripherals and peripherals['rv_plic']['is_included'] == 'yes':
+% if user_peripheral_domain.contains_peripheral('rv_plic'):
```
davidmallasen, line 296: Since now we're not including the signals if the peripheral is not there, this if statement should englobe as well the `reg_to_tlul_plic_i` right above (move the `% if` line right above that block. If not, if the rv_plic is not included, the `core_v_mini_mcu_pkg::RV_PLIC_IDX` is not generated and the build fails.

hw/core-v-mini-mcu/peripheral_subsystem.sv.tpl
```diff
@@ -461,7 +463,7 @@ module peripheral_subsystem
       .reg_rsp_o(peripheral_slv_rsp[core_v_mini_mcu_pkg::RV_TIMER_IDX])
   );
 
-% if 'rv_timer' in peripherals and peripherals['rv_timer']['is_included'] == 'yes':
+% if user_peripheral_domain.contains_peripheral('rv_timer'):
```
davidmallasen, line 466: Same here, this if should also englobe the `rv_timer_reg_to_tlul_i` right above

hw/core-v-mini-mcu/peripheral_subsystem.sv.tpl
```diff
@@ -391,7 +393,7 @@ module peripheral_subsystem
       .reg_rsp_o(peripheral_slv_rsp[core_v_mini_mcu_pkg::I2C_IDX])
   );
 
-% if 'i2c' in peripherals and peripherals['i2c']['is_included'] == 'yes':
+% if user_peripheral_domain.contains_peripheral('i2c'):
```
davidmallasen, line 396: Same here, this if should include the `reg_to_tlul_i2c_i` above

util/x_heep_gen/load_config.py
```diff
@@ -179,6 +208,178 @@ def load_linker_config(system: XHeep, config: list):
...
+                            num_channels=int(peripheral_config["num_channels"], 16),
+                            num_master_ports=int(
+                                peripheral_config["num_master_ports"], 16
+                            ),
+                            num_channels_per_master_port=int(
+                                peripheral_config["num_channels_per_master_port"], 16
+                            ),
+                            fifo_depth=int(peripheral_config["fifo_depth"], 16),
+                            addr_mode=(
+                                0 if peripheral_config["addr_mode_en"] == "no" else 1
```
davidmallasen, line 282: Here you are passing the `addr_mode` as 0 or 1 and then in the DMA class constructor you are expecting it as "yes" or "no". Adapt it in the constructor (and/or here), and rais a ValueError if you get an unexpected value (to avoid this happening again in the future)

davidmallasen (changes requested):

davidmallasen:
Remember to also delete the `util/dma_subsystem_gen.py` as we discussed

util/x_heep_gen/peripherals/abstractions.py
Pacsort17: This was added because some config files are generated from template files, and thus the first if fails the first time you type make mcu-gen (and the custom python config). I don't really understand why it fails, since it checks if there is either a template to generate the config file, or if the config file exists.
I can comment that, but when the peripheral will be generated, the config file will not exist and there will be further problems

davidmallasen:
We also have a bigger problem that we hadn't spotted. If you:
1. Set i2s to not included in `mcu_cfg.hjson`
2. Run `make mcu-gen`
3. Run `make app`
You will see that the sw drivers rely on having some of the variables set for all of the peripherals (in this case you'll see an error for the I2S_START_ADDRESS). There is no easy solution as for the hw case that I tell you above. In this case, the easiest solution that we can take here is to always set those *_START_ADDRESS variables in `sw/device/lib/runtime/core_v_mini_mcu.h.tpl` even if the user peripheral is not included. For that, you can add some list in the `UserPeripheralDomain` and have a function that returns the name of all of the user peripherals that are not included. Then, in `core_v_mini_mcu.h.tpl`, you set to 0 the START_ADDRESS of those peripherals that are not included. Let me know if you have any questions.

Pacsort17:
> We also have a bigger problem that we hadn't spotted. If you:
> 
>     1. Set i2s to not included in `mcu_cfg.hjson`
> 
>     2. Run `make mcu-gen`
> 
>     3. Run `make app`
>        You will see that the sw drivers rely on having some of the variables set for all of the peripherals (in this case you'll see an error for the I2S_START_ADDRESS). There is no easy solution as for the hw case that I tell you above. In this case, the easiest solution that we can take here is to always set those *_START_ADDRESS variables in `sw/device/lib/runtime/core_v_mini_mcu.h.tpl` even if the user peripheral is not included. For that, you can add some list in the `UserPeripheralDomain` and have a function that returns the name of all of the user peripherals that are not included. Then, in `core_v_mini_mcu.h.tpl`, you set to 0 the START_ADDRESS of those peripherals that are not included. Let me know if you have any questions.

I don't really like this solution, it feels like it will be hard to remove and/or maintain later. Instead, I only hardcoded that in sw/device/lib/runtime/core_v_mini_mcu.h.tpl, thus if you add a peripheral you will have only one change to do (coming in the next commit). Furthermore, testing everything for all peripherals takes forever. Is their an easier way to test things without simulating all ?

util/x_heep_gen/peripherals/abstractions.py
Pacsort17: I tried again the issue I had in the past (make clean-all and then make mcu-gen X_HEEP_CFG=configs/example.py) but now everything works, so it must have changed during the semester. 
The issue was that power_manager has only a <config_path>.tpl file, but no <config_path> file (because it was generated from tpl), thus without the second test it was crashing. I will comment it and we will see in the next PR !

util/x_heep_gen/load_config.py
davidmallasen: Nvm, it was quicker to fix it myself and test again

davidmallasen (approved):

PR #708, opened 2025-05-02
Control of the SPI slave MISO output enable

JuanSapriza:
In the spi slave, the output enable was set by default as output. 
This is ok if you only have one master and one slave, but if you want to connect more than one (e.g. HEEPidermis) the slave would be forcing the MISO signal to remain high as long as it is not its turn to speak, making it impossible for other slaves to speak. 

The solution for this is to set it as inout and controlling the output enable with the chip select connected to the spi slave (which is active low). This way the pad is left in high-impedance while it's not the turn of the spi slave. 

I tested this change on verilator and FPGA and it works fine :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -327,7 +327,7 @@ test:
 ## Builds the specified app, loads it into the programmer's flash and then opens picocom to see the output
 ## @param PROJECT=<folder_name_of_the_project_to_be_built>
 run-fpga-flash-load:
-	$(MAKE) app LINKER=flash_load TARGET=pynq-z2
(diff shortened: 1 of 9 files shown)


JuanSapriza:
I am happy with this. On HEEPidermis we will use the spi slave with the spi flash to test them both, as we will not have an spi host. 

Should i do the same for x-heep? I think it's better to leave the spi host as a nice example of having two spis. Now on HEEPidermis i just created a new app that does the SAME as the spi slave example but with the spi flash instead of host :man_shrugging:  I think that's good enough

PR #718, opened 2025-05-23
CIC filter : Allow to change the number of CIC stages and the comb delay D at runtime

jr-moul:
Allow to change the number of CIC stages at runtime, between 1 and a maximum value. This max value correspond to the bit-width of a register field


hw/ip/pdm2pcm/rtl/pdm_core.sv.tpl
```diff
@@ -198,6 +201,7 @@ module pdm_core #(
       .rstn_i(rstn_i),
       .clr_i (s_clr),
       .en_i  (combs_en),
```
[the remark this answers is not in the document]
jr-moul: I'm sorry, which parameter ?

hw/ip/pdm2pcm/rtl/cic_integrators.sv
```diff
@@ -51,5 +52,21 @@ module cic_integrators #(
...
+
+  always_comb begin
+    msb_index = '0;
+    for (int k = STAGES - 1; k >= 0; k--) begin
+      if (par_cic_activated_stages[k]) begin
+        msb_index = k[$clog2(STAGES)-1:0] + 1;
+        break;
+      end
+    end
+  end
```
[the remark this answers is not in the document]
JuanSapriza: that could be true, good!

davidmallasen:
This will clash with #722 @JuanSapriza @jr-moul, just so you know that we'll have to do a merge/rebase

JuanSapriza:
@jr-moul @davidmallasen  I am merging main on this branch now. I resolved the conflicts and will proceed to run some tests

jr-moul (commented):
When checking waveform, I discovered that we didn't sampled at the expected frequency the input, but half of that (because of the way it uses r_send). This change rectify this, by dividing dividing the parameter by two to make the value in C correspond to the actually used sampling frequency (pcm_clk_o).

However, it lead to an inconsistency in the naming, where div_clk is not actually divided by the value inputted from the C, but half of that. The functionality is not affected since everything is still linked to r_send via the enable signal, which now oscillate at the expected frequency from an user perspective.

This was not detected before since we were not sampling real signal, but only reading from a table.

JuanSapriza:
good catch! I will review now.

hw/ip/pdm2pcm/rtl/pdm2pcm.sv.tpl
```diff
@@ -129,16 +129,32 @@ module pdm2pcm #(
...
+  // 
+  // Without this factor, the actual sampling frequency would be incorrectly 
+  // calculated as half the intended value.
+  //
+  // Corrected frequency relationships:
+  //   actual_sampling_frequency = freq(clk_i) / clkdividx
+  //   actual_output_frequency   = freq(clk_i) / (clkdividx * decimcic)
+  
+  localparam integer CLK_DIV_CORRECTION_FACTOR = 1;
+  wire [ClkDivIdxWidth-1:0] corrected_par_clkdiv_idx;
```
JuanSapriza: just change `wire` for `logic`

hw/ip/pdm2pcm/rtl/pdm2pcm.sv.tpl
```diff
@@ -129,16 +129,32 @@ module pdm2pcm #(
...
+      .DIV_VALUE_WIDTH(ClkDivIdxWidth-CLK_DIV_CORRECTION_FACTOR),
       .DEFAULT_DIV_VALUE(2)
   ) clk_int_div_inst (
       .clk_i(clk_i),
       .rst_ni(rst_ni),
       .en_i(reg2hw.control.enabl.q),
       .test_mode_en_i(1'b0),
       .clk_o(div_clk),
-      .div_i(par_clkdiv_idx >> 1),
+      .div_i((corrected_par_clkdiv_idx[ClkDivIdxWidth-CLK_DIV_CORRECTION_FACTOR-1:0])),
```
JuanSapriza: why do you discard the MSB of the variable? I understand it's not used, but maybe it should have been declared with one less bit from the beginning.  

What i find confusing is doing these kind of operations here and not when the values are assigned.

hw/ip/pdm2pcm/rtl/pdm2pcm.sv.tpl
```diff
@@ -129,16 +129,32 @@ module pdm2pcm #(
...
+  //   actual_sampling_frequency = freq(clk_i) / clkdividx
+  //   actual_output_frequency   = freq(clk_i) / (clkdividx * decimcic)
+  
+  localparam integer CLK_DIV_CORRECTION_FACTOR = 1;
+  wire [ClkDivIdxWidth-1:0] corrected_par_clkdiv_idx;
+  assign corrected_par_clkdiv_idx = par_clkdiv_idx >> CLK_DIV_CORRECTION_FACTOR;
+
   clk_int_div #(
-      .DIV_VALUE_WIDTH(ClkDivIdxWidth),
+      .DIV_VALUE_WIDTH(ClkDivIdxWidth-CLK_DIV_CORRECTION_FACTOR),
```
JuanSapriza: Why not have a localparam with "correctedClkDivIdxWidth" as well, instead of doing the change here?

sw/applications/example_cic/main.c
```diff
@@ -50,9 +50,9 @@ int main(int argc, char *argv[])
 
     mmio_region_t pdm2pcm_base_addr = mmio_region_from_addr((uintptr_t)PDM2PCM_START_ADDRESS);
 
-    mmio_region_write32(pdm2pcm_base_addr, PDM2PCM_CLKDIVIDX_REG_OFFSET ,15);
+    mmio_region_write32(pdm2pcm_base_addr, PDM2PCM_CLKDIVIDX_REG_OFFSET, 16); // need to be an even number
```
JuanSapriza, line 53: What happens if we don't follow this restriction? 
I don't think we should care now, but just to think

jr-moul: Since I'm doing a shift by 2 later, it just round the value to the lower even value (if we input 15, the actual sampling frequency will be sys_clk/14)

jr-moul: But if we input 16 it will be sys_clk/16

jr-moul: So it doesn't "break" anything, just it won't correspond. But then, we we are dividing the frequency, it is almost always by a power of 2, so it should not be an issue.

JuanSapriza: exactly! but then you would think you are sampling at sys_clk/15 and you are actually sampling at sys_clk/14... so there is an inconsistency at the sw level. If you then use those values to estimate something (e.g. to set a timer) it will be wrong. 
Nothing to be changed in my opinion, but this is why i preferred to solve it in sw than in hw originally.

jr-moul: I added a comment in the verilog to explain the issue

hw/ip/pdm2pcm/rtl/pdm2pcm.sv.tpl
jr-moul: Because we can't (or I don't know how) do something like :
`assign corrected_par_clkdiv_idx = (par_clkdiv_idx >> 1)[15:0]`
We need to assign the result of the shift to an intermediate signal of the right width, and then "compress" it

hw/ip/pdm2pcm/rtl/pdm2pcm.sv.tpl
jr-moul: done 👍

hw/ip/pdm2pcm/rtl/pdm2pcm.sv.tpl
JuanSapriza: yep... i think that is sadly not possible. The way of doing it is with yet another variable, but for clarity i would add it. After that i will happily test and (request a) merge

PR #720, opened 2025-05-26
xheep-gen improvements

Pacsort17:
Changes in this PR :
 - Avoid computing xheep object multiple times
 - Add Black Formatter
 - Move config files into config and merge them


hw/core-v-mini-mcu/include/core_v_mini_mcu_pkg.sv.tpl
```diff
@@ -145,9 +145,10 @@ package core_v_mini_mcu_pkg;
   // ---------------------
 
   localparam AO_PERIPHERALS = ${len(base_peripheral_domain.get_peripherals())};
-  localparam int DMA_CH_NUM = ${dma.get_num_channels()};
```
davidmallasen, line 148: Why these changes now?

hw/core-v-mini-mcu/peripheral_subsystem.sv.tpl
```diff
@@ -293,7 +293,6 @@ module peripheral_subsystem
       .reg_req_i(peripheral_slv_req[core_v_mini_mcu_pkg::RV_PLIC_IDX]),
       .reg_rsp_o(peripheral_slv_rsp[core_v_mini_mcu_pkg::RV_PLIC_IDX])
   );
-
```
davidmallasen: No need for these whitespace changes

pad_cfg.hjson
davidmallasen, line 1: Remember to separate the pads from the rest of the configuration

test/test_x_heep_gen/template.hjson.tpl
```diff
@@ -77,7 +77,7 @@
     pdm2pcm_cic_only: 0x0
         %endif
     %endif
-
+
```
davidmallasen: Remove

util/git-diff.py
davidmallasen, line 1: Remember to remove the black formatting changes to the Python files that are not in util/x_heep_gen, so `git-diff.py`, `sv2v_in_place.py`, and `vendor.py` should not be modified

Makefile
```diff
@@ -42,11 +42,12 @@ LINKER   ?= on_chip
...
 TARGET   	?= sim
-MCU_CFG_PERIPHERALS  	?= mcu_cfg.hjson
 X_HEEP_CFG  ?= configs/general.hjson
-PAD_CFG  	?= pad_cfg.hjson
-EXT_PAD_CFG ?=
+PYTHON_X_HEEP_CFG ?= 
 
+# Cached xheep object location
+XHEEP_CACHE ?= build/xheep_cache.pickle
+CACHED = --cached_path $(XHEEP_CACHE) -ca
```
davidmallasen: I think it's clearer if you remove this CACHED shortcut and leave those flags explicitely in the mcu_gen calls below. Also, keep the long version of it so it's more readable (cached instead of ca)

util/x_heep_gen/load_config.py
```diff
@@ -290,10 +290,10 @@ def load_peripherals_config(system: XHeep, config_path: str):
                                 peripheral_config["num_channels_per_master_port"], 16
                             ),
                             fifo_depth=int(peripheral_config["fifo_depth"], 16),
-                            addr_mode=addr_mode_en,
-                            subaddr_mode=subaddr_mode_en,
-                            hw_fifo_mode=hw_fifo_mode_en,
-                            zero_padding=zero_padding_en,
+                            addr_mode=(0 if addr_mode_en == "no" else 1),
```
davidmallasen: Why these changes again? Didn't we have it working correctly now?

util/x_heep_gen/peripherals/abstractions.py
```diff
@@ -419,7 +418,6 @@ def validate(self):
         """
 
         return (
-            self.__check_paths()
```
davidmallasen: This was also working correctly now, I don't think there's need to change

hw/core-v-mini-mcu/peripheral_subsystem.sv.tpl
Pacsort17: that's just when reformating files, I think it was done automatically

Makefile
```diff
@@ -165,6 +167,11 @@ mcu-gen-help:
 verible:
 	util/format-verible;
 
+## Runs black formating for python xheep generator files
+black:
```
davidmallasen: I think the name that I suggested is easier to remember by people who don't know what black is. Also instead of using black directly, calling it with the $PYTHON variable makes sure we're using the correct one:
~~~
format-python:
    $(PYTHON) -m black util/x_heep_gen
~~~

davidmallasen:
I added a few small changes myself. I think it should be good to go now, yes. I'll test further internally and then merge

PR #739, opened 2025-08-15
Mild cleanup of some top-level files

davidmallasen:
This is an attempt to declutter a bit the top level of the repo. Any suggestions are welcome.

I also want to move the .upf files, since they can easily confuse new people. Where would it be a good place to put them? @davideschiavone @StMiky @JuanSapriza

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 <br />
 <p align="center"><img src="docs/source/images/x-heep-outline.png" width="500"></p>
 
+# X-HEEP
+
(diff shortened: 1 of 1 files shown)


davideschiavone:
Please check the UPF flow as the file is expected to be where it was https://github.com/esl-epfl/x-heep/blob/2e175dc1b8f44108d3e82946924e49d5c548eadf/scripts/sim/modelsim/patch_modelsim_Makefile.py#L44

PR #743, opened 2025-09-11
porting the flash model to verilator [single-mode read only]

davideschiavone:


docs/source/How_to/Simulate.md
```diff
@@ -110,12 +110,7 @@ You can also use vopt for HDL optimized compilation:
 make questasim-sim-opt
 ```
 
-then go to
```
davidmallasen, line 113: Why remove this @davideschiavone? The run target should be in the build folder as generated by fusesoc

sw/applications/example_spi_read/main.c
```diff
@@ -19,9 +19,9 @@
 #define PRINTF_IN_SIM   0
 
 #if TARGET_SIM && PRINTF_IN_SIM
-        #define PRINTF(fmt, ...)    printf(fmt, ## __VA_ARGS__)
```
davidmallasen: These should stay as lowercase printf, no?

docs/source/How_to/Simulate.md
davideschiavone: sim_opt is not a fusesoc target

davidmallasen: Then it might be just sim. I think I fixed it in my other draft PR

sw/applications/example_spi_read/main.c
davideschiavone: I didnt touch this

docs/source/How_to/Simulate.md
davideschiavone: yes but as you are gonna do the build maketarget and remove then non-opt, I wouldnt care in this PR

sw/applications/example_spi_read/main.c
davideschiavone: oh yes I did, my bad

PR #746, opened 2025-09-23
Fix CMakeLists regex to match "main"

cousteaulecommandant:
The MATCHES regex that matches and skips source files named "main" is incorrect and actually matches any file containing "main", such as "remain.c" or "maintain.c".

There were two issues in the MATCHES pattern:
1. The `.*` in `/.*${MAINFILE}` (which I guess was added to include both `/main` and `/.*/main`) will also match any string at the beginning of the file name, e.g. `/remain`.
2. The `\` in `"...${MAINFILE}\."` needs to be escaped as `\\` because of the double quotes (`"\\."` becomes `\.` which is the regex for a literal `.`, but `"\."` becomes just `.` which is the regex for "match any character"), so `"${MAINFILE}\."` will match the `main.` in `main.c` but also the `maint` in `maintain.c`.

### DIFF
--- a/sw/CMakeLists.txt
+++ b/sw/CMakeLists.txt
@@ -145,7 +145,7 @@ FOREACH(file_path IN LISTS new_list)
     endif()
   elseif( (${file_path} MATCHES "${SOURCE_PATH}external/") AND ( NOT  ${file_path} MATCHES "exclude" ) )
     SET(add 1)
-  elseif( ${file_path} MATCHES "${SOURCE_PATH}applications/${PROJECT}/.*${MAINFILE}\." ) # look for main.*
+  elseif( ${file_path} MATCHES "${SOURCE_PATH}applications/${PROJECT}/${MAINFILE}\\." ) # look for main.*
     SET(app_found 1)
   elseif( ${file_path} MATCHES "${SOURCE_PATH}applications/${PROJECT}/" ) # other sources
     SET(add 1)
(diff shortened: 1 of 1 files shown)


cousteaulecommandant:
Actually, should the regex check for "main" files within a subdirectory of the project directory at all, or only at its top?  The pattern to find the "main" file is something like `${PROJECT}/${MAINFILE}.c` and it doesn't search within subdirectories, so one could have a project that looks like:
```
sw/applications/project/
|-- main.c   <-- this is the "real" main
|-- tools.c
|-- tools.h
'-- examples/
    |-- dma/
    |   '-- main.c, dma.c, dma.h
    '-- uart/
        '-- main.c
```
with multiple "main.c" files, which are not the project's main file but just happen to be named "main".

If this is a possibility, then I need to remove the `(.*/)?` part of the regex (which I added to substitute the `.*` in the original because I assumed that's what it was for, so that you could put the `main.c` file wherever you wanted, but that doesn't seem to be the case).

davidmallasen:
I have no strong opinion on this. If we remove the regex, we are forcing the user to have the main.c in the top directory of the project. If we leave it, we don't allow multiple main.c files. I slightly favor the second option (leaving the fixed regex), but just because I see less the use case of having multiple main.c files.

sw/CMakeLists.txt
```diff
@@ -145,7 +145,7 @@ FOREACH(file_path IN LISTS new_list)
     endif()
   elseif( (${file_path} MATCHES "${SOURCE_PATH}external/") AND ( NOT  ${file_path} MATCHES "exclude" ) )
     SET(add 1)
-  elseif( ${file_path} MATCHES "${SOURCE_PATH}applications/${PROJECT}/.*${MAINFILE}\." ) # look for main.*
+  elseif( ${file_path} MATCHES "${SOURCE_PATH}applications/${PROJECT}/(.*/)?${MAINFILE}\\." ) # look for main.*
```
JuanSapriza: Hello! Sorry for the latency. I am with @cousteaulecommandant on this, I believe it's better to force main to be on the base folder of the application (i.e. i would remove the `(.*/)?` )

JuanSapriza (changes requested):

sw/CMakeLists.txt
cousteaulecommandant: Done!

PR #752, opened 2025-10-06
Add support for Verilator 5.040

StMiky:
## Description
This PR makes a bunch of changes to make X-HEEP compatible with Verilator 5.040 while trying not to break backwards compatibility with Verilator 4.X. These is achieved through different strategies:
- Directly modify the RTL owned by ESL (e.g., `cv32e40px`)
- Patch vendored files triggering errors in Verilator 5 (e.g., delay statements like `#0.1`)
- Waive warnings supported by both Verilator 4 and 5, but only triggered in the latter
- Conditionally add Verilator 5 specific waiver files for warnings not supported by Verilator 4.

The last point is achieved through a FuseSoC generator declared in `core-v-mini-mcu.core`, implemented in `util/waiver-gen.py`, and called in other `.core` files (e.g., `verilator_waiver.core`) when exporting the dependency tree in the `default` target. The generator automatically detects the Verilator version from the executing environment. If Verilator 5 is detected, the dedicated waiver files are exported together with the old ones.

Because backwards compatibility comes at the cost of increased maintenance effort and missing out on new features, it may be a good idea to simply _discontinue support for Verilator 4.210_ in future releases.

### Pros
- RTL simulation is about 2.5 times faster
- X-HEEP is backward-compatible with Verilator 4

### Cons (due to backward compatibility)
- Verilator 5 waiver statements must be put in dedicated files, e.g., `hw/vendor/waiver/lint/common_cells_v5.vlt`, which increases the maintenance effort.
- Useful v5 features like `--timing` cannot be used while maintaining backwards compatibility, unless different `.core` files are used (e.g., with an automatically-generated top-level `.core` file).

## Status
- [x] Support Verilator 5
- [x] Maintain backwards compatibility with Verilator 4.210
- [x] Update the docs
- [ ] ~Update the CI Docker container with Verilator 5.X~ (this is coming to a Docker/CI-specific PR)


davidmallasen:
Can you add the `util/waiver-gen.py` to https://github.com/esl-epfl/x-heep/blob/4020495f8ffcafebbcf5720a81ad54e8d1b4e9f0/.github/workflows/lint.yml#L72 so that we also use the black formatter here? I didn't add it to every legacy python file yet, but we can keep it consistent with the new ones for now

StMiky:
> Can you add the `util/waiver-gen.py` to
> 
> https://github.com/esl-epfl/x-heep/blob/4020495f8ffcafebbcf5720a81ad54e8d1b4e9f0/.github/workflows/lint.yml#L72

Done and formatted 👍

core-v-mini-mcu.core
```diff
@@ -320,6 +324,8 @@ targets:
         - ams_sim? (pre_patch_vcs_ams_Makefile)
         - tool_xcelium? (pre_build_uartdpi)
         - tool_xcelium? (pre_build_remote_bitbang)
+      post_run:
+        - tool_verilator ? (print_uart_log)
```
StMiky: @davidmallasen We could enable this for other tools as well actually

core-v-mini-mcu.core
```diff
@@ -376,7 +382,7 @@ targets:
           - '--x-assign unique'
           - '--x-initial unique'
           - '--exe tb_top.cpp'
-          - '-CFLAGS "-std=c++11 -Wall -g -fpermissive"'
+          - '-CFLAGS "-Wall -fpermissive"'
```
StMiky, line 385: Removing `-g` makes build and simulation faster.
Verilator 5 requires G++14 or newer, so I removed the `-std` directive

core-v-mini-mcu.core
```diff
@@ -410,7 +416,7 @@ targets:
           - '--x-assign unique'
           - '--x-initial unique'
           - '--exe tb_sc_top.cpp'
-          - '-CFLAGS "-std=c++11 -Wall -g -fpermissive"'
+          - '-CFLAGS "-std=c++14 -Wall -fpermissive"'
```
StMiky: @davideschiavone The SystemC TB build fails due to illegal type conversions. This may be the cause.

hw/vendor/patches/pulp_platform_pulpissimo/SimJTAG.sv.patch
```diff
@@ -0,0 +1,44 @@
+diff --git a/rtl/tb/SimJTAG.sv b/rtl/tb/SimJTAG.sv
+index 7c6c1101..0fce7015 100644
+--- a/rtl/tb/SimJTAG.sv
++++ b/rtl/tb/SimJTAG.sv
+@@ -1,5 +1,12 @@
+ // See LICENSE.SiFive for license details.
+ //VCS coverage exclude_file
++
++`ifdef VERILATOR
++`define ASSIGN_DELAY
```
StMiky, line 10: Verialtor 5 supports delay statements, but the `--timing` or `--no-timing` options shall be specified. However, these options would break backward compatibility with Verilator 4, so I went for `ifdef` instead.

tb/tb_sc_top.cpp
```diff
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
 
 #include "verilated.h"
-#include <verilated_vcd_sc.h>
+#include <verilated_fst_sc.h>
```
StMiky, line 6: I changed this to FST for consistency with the C++ TB

tb/testharness.sv
```diff
@@ -15,11 +15,20 @@ module testharness #(
     parameter USE_EXTERNAL_DEVICE_EXAMPLE = 1,
     parameter CLK_FREQUENCY               = 'd100_000  //KHz
 ) (
-    inout wire clk_i,
-    inout wire rst_ni,
-
-    inout wire boot_select_i,
-    inout wire execute_from_flash_i,
+`ifdef VERILATOR
```
StMiky, line 18: Verilator 5 supports tri-state buffers in the top-level interface, but that requires the `--pins-inout-enables` option and a modified TB that would break backward compatibility, therefore this hack.

tb/testharness.sv
```diff
@@ -341,53 +378,19 @@ module testharness #(
...
     end
   end
 
-  assign delayed_tb_cpu_subsystem_powergate_switch_ack_n = tb_cpu_subsystem_powergate_switch_ack_n[SWITCH_ACK_LATENCY];
-  assign delayed_tb_peripheral_subsystem_powergate_switch_ack_n = tb_peripheral_subsystem_powergate_switch_ack_n[SWITCH_ACK_LATENCY];
-  assign delayed_tb_memory_subsystem_banks_powergate_switch_ack_n = tb_memory_subsystem_banks_powergate_switch_ack_n[SWITCH_ACK_LATENCY];
-  assign delayed_tb_external_subsystem_powergate_switch_ack_n = tb_external_subsystem_powergate_switch_ack_n[SWITCH_ACK_LATENCY];
-
-  always_comb begin
-`ifndef VERILATOR
```
StMiky, line 377: Forcing internal signals triggered compilation errors. I exposed these signals in outside of the `x_heep_system` which is not ideal, so let's discuss about this.

Makefile
```diff
@@ -225,59 +226,52 @@ questasim-build-opt-upf: questasim-build
...
-	cd ./build/openhwgroup.org_systems_core-v-mini-mcu_0/sim-verilator; \
-	./Vtestharness +firmware=../../../sw/build/main.hex $(SIM_ARGS); \
-	cat uart0.log; \
-	echo '<end of uart0.log>'; \
-	cd ../../..;
+	$(FUSESOC) --cores-root . run --no-export --target=sim --tool=verilator $(FUSESOC_FLAGS) --run openhwgroup.org:systems:core-v-mini-mcu $(FUSESOC_PARAM) \
+		--run_options="+firmware=../../../sw/build/main.hex $(SIM_ARGS)"
 
-## Uses verilator to simulate the HW model and run the FW
+## First builds the `PROJECT` app and then launches the RTL simulation with the
```
davidmallasen: This target doesn't build the app, it's the previous one that already has the comment

README.md
```diff
@@ -10,7 +10,7 @@ so that you can extend it with your own accelerator without modifying the MCU, b
 By doing so, you inherit an IP capable of booting RTOS (such as `freeRTOS`) with the whole FW stack, including `HAL` drivers and `SDK`,
 and you can focus on building your special HW supported by the microcontroller.
 
-`X-HEEP` supports simulation with Verilator, Questasim, etc. Morever, FW can be built and linked by using `CMake` either with gcc or with clang. It can be implemented on FPGA, and it supports implementation in Silicon, which is its main (but not only) target. See below for more details.
+`X-HEEP` currently supports simulation with Verilator, Questasim, VCS, and Xcelium. Morever, the firmware can be built and linked using `CMake` with either _GCC_ or _CLang_ as backends. It can be implemented on FPGA, and it supports ASIC implementation in silicon, which is its main (but not the only) target for the platform. See below for more details.
```
davidmallasen: _CLang_ -> _Clang_

core-v-mini-mcu.core
davidmallasen: I agree. This could be there after _every_ simulation, right? Just printing the UART output. I guess it's just removing the `tool_verilator ?`

core-v-mini-mcu.core
davidmallasen: Yeah, I guess we'd only need the `-g` for debugging (which we could add manually)

core-v-mini-mcu.core
davidmallasen: You don't remove the `-std` flag here?

docs/source/GettingStarted/Setup.md
```diff
@@ -114,16 +114,15 @@ cmake --build . --target install
...
 ### 4. Install Verilator:
 
-Verilator requires the [following packages](https://verilator.org/guide/latest/install.html) to be installed (Check [OS requirements](#1-os-requirements) for Ubuntu distribution). The [documentation](https://verilator.org/guide/latest/install.html) page contains instructions for other linux distributions. 
+X-HEEP supports Verilator version 5.040, which requires the [following packages](https://verilator.org/guide/latest/install.html) to be installed (Check [OS requirements](#1-os-requirements) for Ubuntu distribution). The [documentation](https://verilator.org/guide/latest/install.html) page contains instructions for other linux distributions. 
 
-```{note}
-Verilator 4.210 will not build with GCC 12.0 or later, so it will need to be built with an older toolchain.
-```
+> [!Note]
+> Backward compatibility with Verilator 4.210 is currently maintained, yet _this is very likely to change_ in future releases, so we strongly suggest against using it. Also, Verilator 4.210 _requires GCC 12.0 or older_, so make sure to configure your environment accordingly if you choose to use it anyway.
```
davidmallasen: _requires GCC older than 12.0_ instead of _requires GCC 12.0 or older_

hw/system/x_heep_system.sv.tpl
```diff
@@ -115,11 +115,6 @@ ${pad.x_heep_system_interface}
 ${pad.internal_signals}
 % endfor
 
-`ifdef FPGA_SYNTHESIS
```
davidmallasen, line 118: Does removing this not break the FPGA flow?

docs/source/GettingStarted/Setup.md
StMiky: Thanks

hw/system/x_heep_system.sv.tpl
StMiky: Those signals are now routed to the top-level interface `x_heep_system` and connected in the simulation or FPGA wrapper. The trigger for this change was a race condition when overriding these signals from the TB in RTL simulation, but it's also justified by having platform-related connections outside of `x_heep_system`.

Makefile
StMiky: Comment changed

core-v-mini-mcu.core
StMiky: Done

core-v-mini-mcu.core
StMiky: Let's double check this with @davideschiavone first

tb/testharness.sv
```diff
@@ -200,7 +223,17 @@ module testharness #(
   end
 `endif
 
-  // X-HEEP system
+  // -------------
+  // X-HEEP SYSTEM
+  // -------------
+  // Inout pins assignments
```
StMiky, line 229: Comment why this

core-v-mini-mcu.core
StMiky: Solved

davidmallasen (approved):

PR #676, opened 2025-03-17
Support ohw compiler

LuigiGiuffrida98:
This PR is to update X-HEEP software to support ohw prebuilt compiler (https://embecosm.com/downloads/tool-chain-downloads/#core-v-top-of-tree-compilers)

:hammer:  TODO:
- [x] Update docker image with the compiler built by embecosm
- [x] Update CI to use the new docker
- [x] GCC - Compile with `on_chip` linker
- [x] GCC - Compile with `flash_load` linker
- [x] GCC - Compile with `flash_exec` linker
- [x] GCC - Run in simulation
- [ ] GCC - Run on FPGA
- [x] Clang - Compile with `on_chip` linker
- [x] Clang - Compile with `flash_load` linker
- [x] Clang - Compile with `flash_exec` linker
- [x] Clang - Run in simulation
- [ ] Clang - Run on FPGA

PR #756, opened 2025-10-11
Add docker workflow, updated tools, and versioning 

LuigiGiuffrida98:
## Description
Add the flow to build the X-HEEP Docker container (by @LuigiGiuffrida98) directly inside the X-HEEP repository and sets up workloads to deploy it, supporting versioning.

## Features breakdown
The PR brings several changes and improvements to the software and hardware build flows, that are summarized below.

### 1. Workflow to build and publish X-HEEP tools
The PR introduces a _Create X-HEEP Release_ GitHub workflow to prepare and publish the tools that are required to experiment X-HEEP and that support its CI:
- The combined GCC and Clang/LLVM toolchains, compiled with the proper flags to support X-HEEP and packaged together for convenience
- An all-in-one, ready-to-use Docker container providing the full environment that is necessary to use X-HEEP (Conda environment, toolchains, Verilator, Verible, etc.)

The workflow is defined in [`.github/workflows/create-release.yml`](https://github.com/LuigiGiuffrida98/x-heep/blob/add-docker/.github/workflows/create-release.yml). This workflow is not hooked to any event, and shall therefore be dispatched _manually_ from the Actions tab on GitHub. When launched, the workflow does the following:
1. Build GCC and LLVM on a `ubuntu-22.04` and package them in a unique tarball that the users or the CI can download and use (if they run a compatible distro).
2. If the toolchain build succeeds, create a `release/v*.*.*` branch (forked from the branch the workflow is launched from) and open a PR to `main`.
3. Prepare a draft release `v*.*.*` with the source code and the toolchain tarball prepared in point 1.
4. Build the X-HEEP Docker image using the toolchain built in point 1, plus the CORE-V GCC toolchain, and publish it to the [GitHub Container Registry (GHCR)](https://github.com/esl-epfl/x-heep/pkgs/container/x-heep%2Fx-heep-toolchain). The image is tagged with `v*.*.*`, but not with `latest` (see next section) not to replace the _stable_ container (i.e., the one corresponding to the revision in `main`).

The purpose of points 2 and 3 is explained in the following section.

### 2. Versioning
To keep the distributed Docker container aligned with different revisions of X-HEEP (e.g., older revisions used in chips), this PR introduces a basic versioning system. The _Create X-HEEP Release_ workflow accepts a `release_tag` parameter, typically in the SEMVER format `vX.Y.Z` (semantic versioning). This tag shall be set to the desired version number (e.g., `v1.0.0`) when launching the workflow. As explained in the previous section, this results in a new PR, called _Release: `v*.*.*`_ being automatically created, together with a draft release (Releases section on GitHub). The PR includes a commit to change the version number in some relevant files, namely:
- The `name: openhwgroup.org:systems:core-v-mini-mcu:*.*.*` string inside `core-v-mini-mcu.core`.
- The `ARG toolchain_version=*.*.*` string inside `util/docker/dockerfile`, used to build the container locally.
It would be nice to also change the docs version, but I did not implement this so far.

If the PR is accepted, the _draft_ specifier is removed from the release, that becomes the `latest` release. Similarly, the Docker image published to the registry by the workflow is tagged `latest`. From that moment onward, the CI running on `main` uses the latest version of the Docker image. 

> [!Note]
> If a user wishes to pull the container locally, black magic is employed to ensure that the correct image is pulled according to the release his clone points to. Similarly, branches not up to date with main still use the correct Docker image for their CI, so breaking changes in the tools/environment do not affect users working on older revisions (as it happens now).

### 3. Updated Tools
The PR brings Verilator 5.040 (#752 ) to the Docker image. In addition, it bumps tool releases as follows:
- Verible: `v0.0-1824-ga3b5bedf` > `v0.0-4023-gc1271a00` (supports more, common statements)
- GCC: `2022.01.17` > `2023.01.03` (updated RISC-V specs)
- Clang/LLVM: `llvmorg-14.0.0` > `llvmorg-19.1.4` (updated RISC-V specs and linking support)

Moving to a newer version of Clang/LLVM allows tofully build the application with LLVM instead of relying on GCC for linking. A side effect is that no app is blacklisted from Clang/LLVM compilation.

> [!Note]
> Actually, the previous `CMakeLists.txt` used black magic to defined a lasagna of spaghetti-like combination of flags and overrides to only pass the applications files (i.e., `main.c` and other sources in the app directory) to the compiler driver in the `cmake` _compile_ stage, while all the other source files (drivers, runtime, `crt0.S`, etc.) were passed, again, to the compiler driver (**not the linker**) in `cmake` _linking_ stage, basically defying the purpose of using `cmake`. This PR brings order in the galaxy and separate the SW build process as it's meant to be.

All the apps (except #779 ) were adapted to work with the new tools and build flow.

Another side effect of moving to newer toolchains is that they use a newer version of the RISC-V specification where CSR instructions were moved into a dedicated extension (Zicsr). Therefore, this PR modifies the default ISA string from `rv32imc` to `rv32imc_zicsr`.

### 4. Updated CI workflow
This PR modifies the CI workflow to leverage the new Docker image, toolchains, and versioning system. In summary, here are the major changes:
- Always use the correct revision of the Docker image, depending on the branch/PR branch
- Compile and link all apps with both GCC and Clang
- Run the RTL simulation on Verilator using the GCC-built apps

One limitation is that the CI does not test the CORE-V toolchain. This requires to edit the CI MCU configuration to use `cv32e40p[x]` as the CPU and modify the ISA string as done by the `init_corev` function in the docker container. As agreed with @davidmallasen , this is coming to a future release.

### 5. Docs
The docs was updated to reflect all of the above changes and have the user result in the same environment used by the CI, regardless of him using the Docker image or the manual setup.

## Dependencies
- depends on #752

## Status
- [x] Update build scripts
- [x] Update tool versions
- [x] Make regression test jobs use the updated container
- [x] Integrate Docker documentation (README.md) inside X-HEEP docs
- [x] Compile and link with LLVM when `COMPILER=clang`
- [x] Fix apps failing with the new toolchains
- [x] Add CI documentation

### DIFF
--- a/sw/CMakeLists.txt
+++ b/sw/CMakeLists.txt
@@ -481,7 +481,6 @@ endif()
 # Specify that we want to link with GCC even if we are compiling with clang
 if (${COMPILER} MATCHES "clang")
   # Call clang here as the flags below are actually meant for a compiler driver, not a linker.
-  # TODO: this part shall call the linker only, leaving compiler flags to the compiler driver only.
   set(CLANG_LINKER_EXE "ld.lld")
 	if( ${PROJECT} MATCHES "freertos" )
 		set( CMAKE_C_LINK_EXECUTABLE "${CLANG_LINKER_EXE} ${CMAKE_EXE_LINKER_FLAGS} \


StMiky:
> @StMiky [this PR](https://github.com/esl-epfl/x-heep/pull/676) already updated the applications to use the compiler by embecosm (openHW) with PULP extensions, I don't know if this is what you want to do with the 4th TODO.

Hi @LuigiGiuffrida98 , thanks for pointing that out. Our idea for todo 4 was to add at least one test with the PULP extensions to the CI, therefore using the CORE-V compiler. I agree that we should do that in your PR, so I'm removing that action point here.

To reduce the tool support effort, simplify the CI, and keep the Docker container smaller, we could ditch the standard compiler altogether, and rely solely on the CORE-V GCC (besides LLVM/Clang), just changing the ISA string according to the CPU configuration since it's multilib. If you agree, I'd merge this one as is (after cleaning), and move the discussion/work about the CORE-V toolchain to the other PR.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,139 @@
+name: CI
+
+on:
+  push:
+    branches: [ "**" ]
+  pull_request:
+    branches: [ "main" ]
```
davidmallasen: We probably want this CI to be triggered always, at any pull request to any branch

configs/ci.hjson
```diff
@@ -13,11 +13,11 @@
             name: code
             start: 0
             #minimum size for freeRTOS and clang
-            size: 0x00000E800
+            size: 0x000018000
```
davidmallasen, line 16: Do we need 3 32kB banks for code?

util/docker/env.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+# Activate the conda environment
+source /opt/conda/etc/profile.d/conda.sh
+conda activate core-v-mini-mcu
+
+# Functions to initialize SW build environment
+# NOTE: we don't use aliases as they would not be picked up by the non-interactive
+#       shell used by GitHub CI.
+init_gcc() {
```
davidmallasen, line 10: I thought we didn't need this any more. If we do (I guess it's for convenience of the docker user?), we should document it somewhere

util/docker/makefile
```diff
@@ -0,0 +1,24 @@
+ROOT_DIR 			:= $(shell git rev-parse --show-toplevel)
+TAG 				?= $(shell git describe --tags --abbrev=0)
+GITHUB_REPOSITORY 	:= ghcr.io/esl-epfl/x-heep
+IMAGE_NAME 			:= x-heep-toolchain
+IMAGE_ID 			:= $(GITHUB_REPOSITORY)/$(IMAGE_NAME)
+
+.PHONY: docker-build docker-push docker-run docker-pull
```
davidmallasen: docker-push doesn't exist anymore. Maybe you can put the phonies individually on top of each target, although this is something very minor

docs/source/GettingStarted/Setup.md
```diff
@@ -4,25 +4,33 @@ There are two ways of setting up X-HEEP. You can either use the provided docker
...
 ## Docker setup
 
-A docker image containing all the required software dependencies is available on [github-packages](https://github.com/orgs/esl-epfl/packages/container/package/x-heep-toolchain).
+A docker image containing all the required software dependencies is available on [github-packages](https://ghcr.io/esl-epfl/x-heep/x-heep-toolchain:latest).
 
-It is only required to install docker and pull the image.
+It is only required to [install Docker](https://docs.docker.com/engine/install/) and pull the image.
 
 ```bash
-docker pull ghcr.io/esl-epfl/x-heep-toolchain:latest
```
davidmallasen, line 12: If we have the docker targets in the makefile, should we just mention those here, so some new user doesn't get confused? Just mentioning the `make -C util/docker docker-pull` I think is less error prone

configs/ci.hjson
StMiky: For some reason, freeRTOS with Clang fails with less code space, even with `-Wl,--gc-sections`

.github/workflows/ci.yml
StMiky: Agreed. I'm changing this to `[ "**" ]`

util/docker/env.sh
StMiky: Yes, that is just a shortcut to set the env variables (`COMPILER`, `COMPILER_PREFIX`, `ARCH`). I'll write a section in the docs.

StMiky: Done.

util/docker/makefile
StMiky: Phonies added, thanks. I intentionally removed `docker-push` as the dedicated _Create X-HEEP Release_ workflow (`.github/workflows/create-release.yml`) shall be preferred.

docs/source/GettingStarted/Setup.md
StMiky: I removed the manual commands in the latest revision

davidmallasen: Here I see that we have both. Using directly `docker pull` and `docker run`, and also using the makefile targets `make -C util/docker docker-pull` and `make -C util/docker docker-run`. I would just leave this second option with the makefile, to avoid confusion with new users

.gitignore
```diff
@@ -72,3 +72,4 @@ flow/OpenROAD-flow-scripts
 .vscode
 .continue
 private/
+GEMINI.md
```
davidmallasen: All good on my side now! I just saw this change while double-checking from top to bottom briefly. Is this needed/intended? Otherwise feel free to merge after solving the conflict

StMiky: I reverted it as that file seems to be picked up even if it's in `private/`

PR #774, opened 2025-10-23
moved uart to peri domain, make other peripheral removables, adding IDs and Conf regs

davideschiavone:
- [x] check uart in peripheral domain works in HW and SW
- [x] check you can remove uart and add assertions in HW
- [x] check you can remove uart and add assertions in SW
- [x] check you can remove other peripherals and add assertions in HW
- [x] check you can remove other peripherals and add assertions in SW
- [x] add XHEEP ID
- [x] test Verilator 4
- [x] test Verilator 5

### DIFF
--- a/hw/core-v-mini-mcu/ao_peripheral_subsystem.sv
+++ b/hw/core-v-mini-mcu/ao_peripheral_subsystem.sv
@@ -27,7 +27,7 @@ module ao_peripheral_subsystem
     output reg_rsp_t [AO_SPC_NUM_RND:0] ao2spc_resp_o,
 
     // SOC CTRL
-    input  logic [31:0] xheep_istance_id_i,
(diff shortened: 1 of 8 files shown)


test/test_apps/test_apps.py
```diff
@@ -208,17 +207,6 @@ def run_app(an_app, simulator):
                 + BColors.ENDC
             )
             print(BColors.FAIL + str(run_output.stdout.decode("utf-8")) + BColors.ENDC)
-
-            # For questasim the build folder is sim-modelsim instead of sim-questasim
-            simulator_build_name = simulator if simulator != "questasim" else "modelsim"
-            uart_output = open(
```
StMiky, line 214: The UART output is now automatically printed by the FuseSoC simulation target. Therefore is already printed with `run_output.stdout` above.

PR #778, opened 2025-10-28
Update mcu gen

davidmallasen:
Modifies several things in the mcu-gen build system:
- Adds extension functionality to mcu-gen, so that it can also be used in systems that extend X-HEEP.
- Redid parameter override (when passing parameters directly to the make target). However, this part could stil be improved in the future. Especially the memory subsystem.
- Refactored and separated into classes both the top level X-HEEP and the memory subsystem.
- Simplified and fixed the configuration loading, both from hjson and python. When using python we still have to pass an hjson with the unsupported configuration (for now).
- Added CPU to the python modelling.
- Smaller improvements

### DIFF
--- a/docs/source/Configuration/PeripheralConfiguration.md
+++ b/docs/source/Configuration/PeripheralConfiguration.md
@@ -18,13 +18,7 @@ Every peripheral has at least an offset and a length, which represent its positi
 
 ## Adding a custom configuration
 
-An example is shown in `configs/general.py` :
-```{literalinclude} ../../../configs/general.py
-:language: python
-:lines: 42-63
-```
-
-First, both domains must be created. If a domain is not created, X-HEEP will be built with the provided HJSON configuration. Base Peripheral domain starts from 0x20000000, User peripheral domain starts from 0x30000000, and each fills 1MB.
+An example is shown in [configs/general.py](https://github.com/esl-epfl/x-heep/blob/main/configs/general.py). First, both domains must be created. If a domain is not created, X-HEEP will be built with the provided HJSON configuration. Base Peripheral domain starts from 0x20000000, User peripheral domain starts from 0x30000000, and each fills 1MB.
 
 Each peripheral has its own class, that must be imported from `x_heep_gen.peripherals.base_peripherals.py`or `x_heep_gen.peripherals.user_peripherals.py`.
 


docs/source/Configuration/ExtensionConfiguration.md
```diff
@@ -0,0 +1,44 @@
...
+
+To add an extension to the X-HEEP system, you can use the `add_extension()` method of the `XHeep` class. For example:
+
+```python
+from x_heep_gen.xheep import XHeep
+
+def config():
+    system = XHeep(BusType.NtoM)
+    ...
+    system.add_extension("my_extension", extension_function())
```
StMiky, line 14: 💯

docs/source/Configuration/SystemConfiguration.md
```diff
@@ -24,40 +31,20 @@ X-HEEP can be configured in more detail using an hjson file. The default configu
...
-- `MEMORY_BANKS_IL` 
-
-These will replace the configuration used in the hjson file. When one parameter is not provided the configuration files value is used.
-The memory banks configured this way will only be 32KB.
-For compatibility reasons `MEMORY_BANKS` does not create linker sections while `MEMORY_BANKS_IL` does create a linker section.
-
-## Python Configuration
+## Python Configuration (recommended)
 
 A more complex configuration can be done using a python script. The default configurations and examples are located in the `configs` directory.
```
StMiky: **P**ython

davidmallasen: What do you mean?

tb/tb_util.svh.tpl
```diff
@@ -2,11 +2,15 @@
 // Solderpad Hardware License, Version 2.1, see LICENSE.md for details.
 // SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
 
+<%
+    memory_ss = xheep.memory_ss()
```
StMiky, line 6: Shall we somehow expose `memory_ss` directly? Just to avoid this in all template files.

davidmallasen: We could add some shortcuts via the kwargs (aka sending the kwarg `xheep` but also the kwarg `memory_ss`). However, I felt it was more structured to always go through the xheep class methods so we don't lose ourselves in the hierarchy, since we can just add these shortcuts once in the files if we want (as you can see in some other files I didn't do the shortcut, either because there was only one instance, or because I didn't trust the `<%` syntax depending on the language). However, if you think it's better to expose some things directly in the kwargs it would be doable. The thing then would be where to draw the line

docs/source/Configuration/SystemConfiguration.md
davidmallasen: I now capitalized all the Pythons in the text of the docs

tb/tb_util.svh.tpl
StMiky: I kind of agree with you about getting lost in the hierarchy matrix. And it's not like we will add that snipped every day in a new file, so yes we can keep the shortcut on a per-file basis.

PR #780, opened 2025-10-29
Fix peripheral structs compilation error

FrancescoPoluzzi:
Add a dummy run of structs_periph_gen generating structs for all peripherals to avoid sw compilation error when a peripheral is not included.
Temporary workaround, should be handled in the sw compilation pipeline.


davidmallasen:
I will fix this in a different PR soon

PR #803, opened 2025-11-10
Allow mcu-gen to use cv32e20 e,m params from python config file

allpaul0:
The command:
make mcu-gen CPU=cv32e20 PYTHON_X_HEEP_CFG="$config_path"
fails when invoked from Python configuration files because mcu-gen does not correctly pickle (serialize) the CPU object using the appropriate class constructor and parameters for cv32e20.

As a result, the serialized cache does not contain a properly instantiated cv32e20 object, causing the subsequent chained mcu-gen call to receive an incomplete CPU instance (instanciated by default).

Fix: Ensure that mcu-gen serializes the CPU using the correct constructor (cv32e20(...)) and its corresponding parameters, so that deserialization restores the correct object type and configuration.

### DIFF
--- a/util/mcu_gen.py
+++ b/util/mcu_gen.py
@@ -19,6 +19,7 @@ import x_heep_gen.load_config
 from x_heep_gen.load_config import load_peripherals_config
 from x_heep_gen.xheep import BusType
 from x_heep_gen.cpu.cpu import CPU
+from x_heep_gen.cpu.cv32e20 import cv32e20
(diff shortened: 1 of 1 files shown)


allpaul0:
Hello, 
I see, this is my mistake, the CPU=cv32e20 discards the subsequent cpu configuration from the hjson or py file.
There is no error in the command execution, I didn't use the command as it was intended.

PR #813, opened 2025-11-17
Fix 'verilator-waves' Makefile target

alfonrod:
The current `Makefile` target to show the output of Verilator-based simulations does not follow the new versioning number and fails.  This PR solves it (although it would probably make more sense to automate it with the automatic release workflow).


davideschiavone:
I think we should do like here: https://github.com/esl-epfl/x-heep/blob/main/Makefile#L218 as otherwise next version of x-heep will have the same issue, but meanwhile thx @alfonrod

davidmallasen:
I agree, that's why I linked the open issue on that

PR #817, opened 2025-11-21
Complete ohw compiler support

LuigiGiuffrida98:
After merging this PR:

- The default compiler in Makefile would be core-v gcc
- In CI the apps would be built with all the supported compilers
- In CI the apps would be simulated with ventilator and built only with the core-v gcc
- Dependencies between CI jobs would be added to avoid running of jobs if some of the basic jobs fails.
Screenshot 2025-11-25 at 15 57 29
- Table mode would be added to test script
Screenshot 2025-12-01 at 12 01 22

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,6 @@ jobs:
     needs: [determine-image-tag, compile-apps]
     container:
       image: ghcr.io/esl-epfl/x-heep/x-heep-toolchain:${{ needs.determine-image-tag.outputs.tag }}
-    if: github.event_name == 'pull_request'
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
--- a/docs/source/GettingStarted/Setup.md
+++ b/docs/source/GettingStarted/Setup.md
@@ -44,7 +44,7 @@ To use `X-HEEP`, first you will need to install some OS dependencies.
 
 The following command `apt` command should install every required package (tested on an Ubuntu 22.04 distribution):
 ```bash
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -55,12 +55,12 @@ jobs:
         shell: bash -l {0}
         run: |
           make mcu-gen X_HEEP_CFG=configs/ci.hjson && \
-          python3 test/test_apps/test_apps.py --compilers gcc,clang --compiler-paths /tools/riscv --compiler-prefixes riscv32-unknown- --compile-only
+          python3 test/test_apps/test_apps.py --compilers gcc,clang --compiler-paths /tools/riscv --compiler-prefixes riscv32-corev-,riscv32-unknown- --compile-only
```
davidmallasen: With this you're already setting the ohw compiler as the default for gcc in the CI. Is this what we want to do?

LuigiGiuffrida98: We can test with the three compilers actually

davidmallasen: Yes, by modifying the testing script, but with this change that I highlight, the CI ran with the ohw compiler only. It depends on what we want to do with this

davidmallasen (approved):

docs/source/GettingStarted/Setup.md
```diff
@@ -18,9 +18,9 @@ The Docker image provides built-in shortcuts (as Bash functions) to select among
 
 | Shortcut | Description | Configuration |
 | -------- | ----------- | ------------- |
+| `init_corev` | Use the Embecosm CORE-V toolchain with PULP extension | `COMPILER=gcc`<br>`COMPILER_PREFIX=riscv32-unknown-`<br>`ARCH=rv32imc_zicsr_zifencei_xcvhwlp_xcvmem_xcvmac_xcvbi_xcvalu_xcvsimd_xcvbitmanip` |
```
davidmallasen: This should be `riscv32-corev-`

PR #818, opened 2025-11-21
Speed up UART in simulation

cousteaulecommandant:
Simulation UART was set to a speed of 256000 baud, when theoretically it can reach up to almost CLK_FREQ/16. This causes UART to be a bottleneck in apps that print a lot of info, making the simulation take very long to finish.

This commit changes the baud rate to CLK_FREQ/20 (close to the limit), i.e. each UART bit lasts exactly 20 clock cycles.
(In other words, baud rate is now 5,000,000.)
The UART_NCO macro has been rewritten to accommodate this new value without overflowing.

Text-heavy applications have been accelerated about a tenfold.

Note that this does not touch the UART speed in FPGA implementation, which still defaults to 9600 baud (could be accelerated to 115200 but this probably won't be an issue).

### DIFF
--- a/sw/device/target/sim/x-heep.h
+++ b/sw/device/target/sim/x-heep.h
@@ -12,15 +12,11 @@ extern "C" {
 #endif  // __cplusplus
 
 
-#define REFERENCE_CLOCK_Hz 100*1000*1000
-#define UART_BAUDRATE 256000
+#define REFERENCE_CLOCK_Hz (100*1000*1000) // 100 MHz for simulation
+#define UART_BAUDRATE (REFERENCE_CLOCK_Hz / 20) // close to maximum baud rate (/16)
 // Calculation formula: NCO = 16 * 2^nco_width * baud / fclk.
-// NCO creates 16x of baudrate. So, in addition to the nco_width,
-// 2^4 should be multiplied.
-// We assume that the lowest baudrate will be 9600, and the largest 256000. 
-// Given this range, by dividing by 100 it always remains integer and below 32-bits.
-// This saves us the need of performing 64-bit divisions to compute NCO. 
-#define UART_NCO ((uint32_t)(((uint32_t)((uint32_t)(UART_BAUDRATE/100))<<20)/((uint32_t)(REFERENCE_CLOCK_Hz/100))))
+// Note that this will be calculated at compile time, so no 64-bit operations will be performed in CPU.
+#define UART_NCO ((uint32_t)( (((uint64_t)UART_BAUDRATE * 16) << 16) / REFERENCE_CLOCK_Hz ))
 #define TARGET_SIM 1
 
 /**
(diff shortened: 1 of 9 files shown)


davidmallasen:
> I see that a new FPGA (ZCU102) was added in the last couple of hours while I was working on this PR, but its x-heep.h doesn't contain the `UART_NCO` macro (probably because the PR is based on an old commit). Should I add it as well?

@LuigiGiuffrida98

LuigiGiuffrida98:
Hi @cousteaulecommandant what is the status of this PR? Can you please rebase it on the main and fix the conflicts?

cousteaulecommandant:
@LuigiGiuffrida98 Done!  Turns out the conflict was because someone else had already applied the same fix you requested me to add to ZCU102 (in a slightly different way as I did so it conflicted).

I have solved the conflicts and also changed all the `sw/device/target/*/x-heep.h` files so that they all use the same formula for `UART_NCO`, and added parentheses to the `REFERENCE_CLK_Hz` macro in all of them.

Please, merge this as soon as possible to avoid further conflicts :)

cousteaulecommandant:
@LuigiGiuffrida98 @davidmallasen All conflicts solved and all CI tests passed.  The PR is ready to merge! 🎉

davidmallasen:
This looks good to me @cousteaulecommandant, thanks! @davideschiavone self-requested a review, so I'll let him check

sw/device/target/sim/x-heep.h
```diff
@@ -12,15 +12,11 @@ extern "C" {
 #endif  // __cplusplus
 
 
-#define REFERENCE_CLOCK_Hz 100*1000*1000
-#define UART_BAUDRATE 256000
+#define REFERENCE_CLOCK_Hz (100*1000*1000) // 100 MHz for simulation
+#define UART_BAUDRATE (REFERENCE_CLOCK_Hz / 20) // close to maximum baud rate (/16)
```
davideschiavone, line 16: I do not understand something here - in this case BAUDRATE is `500000` which is not a standard value - am I wrong?

davideschiavone (changes requested):

sw/device/target/sim/x-heep.h
cousteaulecommandant: @davideschiavone The original goal of this commit was precisely to increase the baudrate to near the highest value supported by the UART peripheral (please refer to the PR title and description, which explains this decision).  So setting the baudrate to 5,000,000 was the entire point of this PR.
Although 5,000,000 is not a *standard* UART baudrate, it is supported by the simulation, which is all that matters since no real hardware is involved, so this is not an issue.  Hardware implementations will still use 9600 as usual.

davideschiavone: true - thx

cousteaulecommandant:
Thanks!

PR #825, opened 2025-11-30
Python Pad Configuration

marinPh:
# Pad Configuration Architecture

  ## Pipeline Stages

  1. **Configuration**: User writes `PadDef` + `PadGroup` (HJSON or Python)
  2. **Build**: `PadRing` computes placement, produces `Pad` objects
  3. **Generation**: Templates use `Pad` objects for RTL/headers

  ## Objects

  **`PadDef`** (configuration-time)
  - Logical pad description: name, type, mux, electrical properties
  - No physical indices or banks, holds layouts that describes physical dimensions

  **`PadGroup`** (configuration-time)
  - Container for `PadDef` instances
  - Global constraints and options
  - Same structure for HJSON and Python

  **`PadRing`** (build-time)
  - Assigns physical positions/indices
  - Validates constraints
  - Outputs `Pad` objects

  **`Pad`** (generation-time)
  - Physical pad for templates: index, bank, position
  - Encoded values for RTL
  - Flat, stable interface

  ## Configuration Formats

  **HJSON**: Declarative, parsed into `PadGroup` + `PadDef`

  **Python** (`pad_cfg.py`): Direct construction with loops/conditionals

  Both produce identical `PadGroup`.

  ## Testing

  1. Generate with HJSON config
  2. Generate with Python config
  3. Compare JSON outputs against golden
  4. Diff to `diff_output.txt` on mismatch

  ## Design

  - **Separation**: Config (`PadDef`/`PadGroup`) → Build (`PadRing`) → Generation (`Pad`)
  - **Single representation**: Both formats use same internal objects
  - **Minimal duplication**: Config has rich data, templates see derived values

### DIFF
--- a/util/x_heep_gen/load_config.py
+++ b/util/x_heep_gen/load_config.py
@@ -1,5 +1,5 @@
 import importlib
-from pathlib import PurePath,Path
+from pathlib import PurePath, Path
 from typing import List, Union
 import hjson
 import os


docs/source/Testing/Testing.md
```diff
@@ -89,13 +125,23 @@ This workflow ensures the stability and integrity of the codebase by running a s
...
+          * repeat for pad_cfg.py
+
 ### Release Workflows
 
 The project includes a robust, automated process for creating and publishing releases. This is handled by two GitHub Actions workflows: `create-release.yml` and `publish-release.yml`. This system ensures that every release is consistently built, tested, and published with its corresponding toolchain and Docker image.
 
 #### Create X-HEEP Release Workflow (`create-release.yml`)
 
-This workflow prepares a new release. It is a comprehensive process that builds the toolchain, packages it, creates a draft release, builds a Docker container, and opens a version bump pull request. It's designed to be triggered manually when a new release is needed.
+This workflow prepares a new release. It is a comprehensive process that builds the toolchain, packages it, creates a draft release, builds a Docker container, and opens a version bump pull request. It`s designed to be triggered manually when a new release is needed.
```
davidmallasen: Revert

util/x_heep_gen/pads/PadDef.py
```diff
@@ -0,0 +1,508 @@
...
+
+class Orientation(Enum):
+    R0 = "R0"
+    R90 = "R90"
+    R180 = "R180"
+    R270 = "R270"
+    MX = "MX"
+    MY = "MY"
+    MX90 = "MX90"
+    MY90 = "MY90"
```
davidmallasen, line 67: Use in the configs

util/mcu_gen.py
```diff
@@ -156,9 +152,10 @@ def generate_xheep(args):
 
     # Here the xheep system is built,
     # The missing gaps are filled, like the missing end address of the data section.
-    xheep.build()
-    pad_ring = PadRing(pad_cfg)
+    if pad_ring is None:
```
davidmallasen: Move this inside the validation of xheep

Makefile
```diff
@@ -121,6 +121,7 @@ conda:
 ## @param X_HEEP_CFG=[configs/general.hjson(default),<path-to-config-file>]
 ## @param PYTHON_X_HEEP_CFG=[configs/general.py(default),<path-to-config-file>]
 mcu-gen:
+	pip install nbformat
```
davidmallasen: Move to python_requirements

util/python-requirements.txt
```diff
@@ -6,6 +6,7 @@ mako==1.1.6
 jsonref==0.2
 black==24.8.0 
 deepdiff
+nbformat
```
davidmallasen: Add a specific version for both this and deepdiff, same as with the other packages

util/x_heep_gen/pads/PadDef.py
```diff
@@ -0,0 +1,504 @@
...
+            # ANSI bright yellow warning
+            warning = (
+                "\033[93m[PadGroup WARNING] One or more physical attributes or "
+                "layout dimensions are missing. All physical properties are "
+                "being set to None.\033[0m"
+            )
+
+            # print all that are missing
+            if self.pad_edge_offset is None:
+                warning += "\n - pad_edge_offset is missing"
```
davidmallasen: Add these in yellow as well

util/x_heep_gen/pads/PadDef.py
```diff
@@ -0,0 +1,504 @@
...
+        pads_cfg = cfg.get("pads", {})
+
+        for pad_name, pad_info in pads_cfg.items():
+            pad_type = PadType(pad_info.get("type", "input"))
+            mapping_str = pad_info.get("mapping", "top")
+            mapping = PadMapping(mapping_str)
+
+            la = pad_info.get("layout_attributes", {})
+            layout_index = la.get("index", 0)
+            print(f"pad_name = {pad_name}, la = {la}")
```
davidmallasen: Is this a debug print?

util/x_heep_gen/pads/PadDef.py
```diff
@@ -0,0 +1,504 @@
...
+from enum import Enum
+
+
+class PadType(Enum):
+    INPUT = "input"
+    OUTPUT = "output"
+    INOUT = "inout"
+    BYPASS_INPUT = "bypass_input"
+    BYPASS_OUTPUT = "bypass_output"
+    BYPASS_INOUT = "bypass_inout"
```
davidmallasen, line 34: We're missing a PadType: `SUPPLY = "supply"`

Check: https://x-heep.readthedocs.io/en/latest/Configuration/PadConfiguration.html#mandatory-pad-attributes

util/x_heep_gen/pads/PadDef.py
```diff
@@ -0,0 +1,504 @@
...
+from dataclasses import dataclass, field
+from typing import Dict, List, Optional, Iterable, Tuple, Mapping, Any
+from enum import Enum
+
+
+class PadType(Enum):
+    INPUT = "input"
+    OUTPUT = "output"
+    INOUT = "inout"
+    BYPASS_INPUT = "bypass_input"
```
davidmallasen, line 32: The types `BYPASS_*` will not have the `layout_attributes`. Therefore, in `util/x_heep_gen/pads/PadRing.py` line 265 I get a `AttributeError: 'NoneType' object has no attribute 'bond_pad'` because the `pad.layout` is None. Also in line 286 the `pad.layout.bond_pad` fails because of the same thing.

Check: https://x-heep.readthedocs.io/en/latest/Configuration/PadConfiguration.html#mandatory-pad-attributes

davidmallasen: This also means that later, the bypass pads will not be phisically in the pad ring, but they will be in the core_v_mini_mcu interface. We can discuss this in person

util/x_heep_gen/pads/PadDef.py
marinPh: yes, will remove before merge

Makefile
```diff
@@ -53,7 +53,7 @@ TARGET   	?= sim
 
 # Mcu-gen configuration files
 X_HEEP_CFG  ?= configs/general.hjson
-PADS_CFG ?= configs/pad_cfg.hjson
```
davidmallasen, line 61: Revert this

docs/source/Configuration/PadConfiguration.md
```diff
@@ -152,3 +152,248 @@ The automatic pad placement calculates the pad `offset` and `skip` parameters su
...
+| `layout_attributes.offset/skip` | `Layout.offset`, `Layout.skip`                              |
+| `physical_attributes.*`         | constructor args of `PadGroup` (`fp_dim`, offsets, spacing) |
+
+---
+
+## Example: Pure Python configuration (`config() -> PadRing`)
+
+A Python pad config file can directly construct the pads and return a `PadRing`:
+
+```python
```
davidmallasen, line 216: Point to the config file

hw/core-v-mini-mcu/core_v_mini_mcu.sv
```diff
@@ -41,10 +41,10 @@ module core_v_mini_mcu
 
     input logic jtag_tdi_i,
 
-    output logic jtag_tdo_o,
```
davidmallasen, line 44: Check why

util/x_heep_gen/pads/PadDef.py
```diff
@@ -0,0 +1,749 @@
...
+        This is the main entry point for loading pad configurations from HJSON.
+        It constructs a PadGroup with all pads and physical properties.
+
+        :param cfg: HJSON configuration dictionary
+        :param str name: Name for the pad group
+        :return: Constructed PadGroup with all pads configured
+        :rtype: PadGroup
+
+        Configuration structure expected:
+            {
```
davidmallasen, line 738: This is missing the attributes section. Check: https://x-heep.readthedocs.io/en/latest/Configuration/PadConfiguration.html#defining-additional-rtl-inputs-for-all-pad-objects

```
{
    attributes: {
        bits: 7:0
        resval: 0x6C
    }
...
}

How is this passing tests with the previous golden reference? This is used in `hw/system/pad_ring.sv.tpl` in lines 16-22 for example.

File: test/test_x_heep_gen/pads/unit/test_pad_positions.py [hunk 14]
```diff
@@ -0,0 +1,313 @@
...
+        mapping=mapping,
+        layout=layout,
+        orient=Orientation.R0,
+    )
+    return pad
+
+
+def test_single_pad_is_centered(sample_pad_group):
+    """Test that a single pad is centered on the side."""
+    # Floorplan is 1000x1000, bondpad edge offset is 20
```
davidmallasen, line 80: Where are these numbers coming from? Are they some defaults? If so, I'd say you shouldn't use defaults here. Same in all the rest.

Makefile
```diff
@@ -346,12 +346,58 @@ test:
 	python3 test/test_x_heep_gen/test_peripherals.py
 	@echo "You can also find the peripheral test outputs in test/test_x_heep_gen/outputs"
 
-.PHONY: test_kwargs
-test_kwargs:
-	$(MAKE) mcu-gen X_HEEP_CFG=configs/ci.hjson PADS_CFG=test/test_x_heep_gen/pads/pad_cfg.hjson
-	$(PYTHON) util/mcu_gen.py --cached_path $(XHEEP_CONFIG_CACHE) --cached --outtpl test/test_x_heep_gen/pads/output/kwargs_output.json.tpl
-	python3 test/test_x_heep_gen/pad_test.py
-
+## Run scenario-based pad configuration tests
```
davidmallasen, line 359: Are we using all of these targets? Should they be all in the top level makefile? If the user only needs a subset of these, we could put them in a dedicated makefile in `tests/test_x_heep_gen`

docs/source/Testing/Testing.md
```diff
@@ -21,7 +21,90 @@ Additionally, you can check only the compilation of the applications with the fo
...
+   - Checks MultiplexedPad mux selector width calculation (e.g., 4 alternatives → 2 bits)
+   - Verifies side-based pad separation (top/bottom/left/right)
+
+**When to use unit tests:**
+- During development of new pad configuration features
+- For quick validation of logic changes
+- To test edge cases and error handling in isolation
+- Before running slower integration tests
+
+### Integration tests (HJSON to Python equivalence)
```
davidmallasen: With all of these additions that you did now, I see a bit of chaos. In the test/test_x_heep_gen: what is the purpose of scenarios, configs, and pads? Why do we have these three sets of configurations? It is very chaotic and hard to understand now.

.gitignore
```diff
@@ -97,5 +99,4 @@ test/test_x_heep_gen/pads/output/kwargs_output.json
 .vscode
 .continue
 private/
-
-
+test/test_x_heep_gen/scenarios/**/output/kwargs_output.json
```
davidmallasen: Don't put ignores in random places here...

.gitignore
```diff
@@ -87,8 +87,10 @@ sw/device/lib/drivers/dma/dma_regs.md
 
 # Openroad
 flow/OpenROAD-flow-scripts
+
 #testing outputs for pads
 test/test_x_heep_gen/pads/output/kwargs_output.json
```
davidmallasen: You have a `.gitignore` in `test/test_x_heep_gen` now. Add these ignores there instead.

.github/workflows/ci.yml
```diff
@@ -126,12 +126,14 @@ jobs:
...
-      - name: Configure the job container and simulate all apps
+      - name: Configure the job container and test pads
         shell: bash -l {0}
         run: |
           make clean-all
           # Run the test script
-          make test_kwargs
+          make test_pads
+          make clean-all
+          make test_pads_unit
```
davidmallasen, line 134: If unit tests are faster, shouldn't these be executed first?

util/python-requirements.txt
davidmallasen, line 1: If you're using pytest explicitly now, should that be included here?

.github/workflows/ci.yml
marinPh: yes

.gitignore
marinPh: yes this is a mistake

docs/source/Testing/Testing.md
marinPh: Scenarios allows to me test multiple configurations continuously, I will clean the others

Makefile
marinPh: I was using them locally but can clean them for the merge

util/python-requirements.txt
marinPh: yes

test/test_x_heep_gen/pads/unit/test_pad_positions.py
marinPh: @pytest.fixture
def sample_pad_group(sample_pad_layout):
    """Create a minimal PadGroup for testing."""
    from x_heep_gen.pads.PadDef import PadGroup, Dimension

    fp_dim = Dimension(width=1000, length=1000)

    return PadGroup(
        name="test_group",
        pad_edge_offset=50,
        bondpad_edge_offset=20,
        bp_spacing=25,
        cell_spacing=None,
        fp_dim=fp_dim,
    )
    
I use this to have a predictable result I can compare

davidmallasen: Ah nice, I didn't know that from pytest

docs/source/Testing/Testing.md
```diff
@@ -57,52 +85,83 @@ Unit tests validate the core logic of the pad configuration framework in isolati
...
-- `max_pads`: Stress test with maximum pad density
-- `tight_spacing`: Minimal spacing between pads
-- `all_edges`: Pads distributed across all 4 chip edges
+The test framework automatically discovers and runs the new scenario.
+
+**Generating golden references:**
+
+```bash
+cd test/test_x_heep_gen/pads
+python3 generate_goldens.py --help              # Show all options
```
davidmallasen: This is different from the makefile now, no?

JuanSapriza:
I leave a comment here, which does not need to be solved now, but if it's not added on this PR we will need to make a second version. 

Offsets and spacings (both between pads and to edges) need more granularity: i.e. each pad should be able to include an offset/skip and each side of the ring should be able to include a different offset to the edges. 


```
(row_margin
  (top
    (io_row ring_number = 1 margin = 0)
    (io_row ring_number = 2 margin = 12.2)
    (io_row ring_number = 3 margin =         29)        <<<<<<<<<<<<
    (io_row ring_number = 4 margin =         100)        <<<<<<<<<<<<
  )
  (bottom
    (io_row ring_number = 1 margin = 0)
    (io_row ring_number = 2 margin = 12.2)
    (io_row ring_number = 3 margin =         24)        <<<<<<<<<<<<
    (io_row ring_number = 4 margin =          95)        <<<<<<<<<<<<
  )
  . . .
)

. . . 

   ( bottom
      (locals
        ring_number = 3
        space = 15
        io_order = clockwise
      )
      ( inst name="u_pad_ring/pad_LDO_VOUT_i_bondpad" cell=xxxx       offset=211       orientation=R180 )        <<<<<<<<<<<<
      ( inst name="u_pad_ring/pad_ACORE_VSS_4_i_bondpad" cell=xxxxx orientation=R180 )
       . . . 
      ( inst name="u_pad_ring/pad_VCON_VIN_i_bondpad" cell=xxxxxD60LA orientation=R180 )
      ( inst name="u_pad_ring/pad_VSS_3_i_bondpad" cell=xxxxx         space=25           orientation=R180 )        <<<<<<<<<<<<
      ( inst name="u_pad_ring/pad_DCORE_VDD_2_i_bondpad" cell=xxxxx      space=15        orientation=R180 )        <<<<<<<<<<<<

```

From the pads side, this can be done on each cheep, modifying the cheep.io.tpl. 
However, from the margins' side, this is hardcoded into the PadRing configuration, if I'm not wrong. It should be as simple as being able to split the margins per side

marinPh:
> I leave a comment here, which does not need to be solved now, but if it's not added on this PR we will need to make a second version.
> 
> Offsets and spacings (both between pads and to edges) need more granularity: i.e. each pad should be able to include an offset/skip and each side of the ring should be able to include a different offset to the edges.
> 
> ```
> (row_margin
>   (top
>     (io_row ring_number = 1 margin = 0)
>     (io_row ring_number = 2 margin = 12.2)
>     (io_row ring_number = 3 margin =         29)        <<<<<<<<<<<<
>     (io_row ring_number = 4 margin =         100)        <<<<<<<<<<<<
>   )
>   (bottom
>     (io_row ring_number = 1 margin = 0)
>     (io_row ring_number = 2 margin = 12.2)
>     (io_row ring_number = 3 margin =         24)        <<<<<<<<<<<<
>     (io_row ring_number = 4 margin =          95)        <<<<<<<<<<<<
>   )
>   . . .
> )
> 
> . . . 
> 
>    ( bottom
>       (locals
>         ring_number = 3
>         space = 15
>         io_order = clockwise
>       )
>       ( inst name="u_pad_ring/pad_LDO_VOUT_i_bondpad" cell=xxxx       offset=211       orientation=R180 )        <<<<<<<<<<<<
>       ( inst name="u_pad_ring/pad_ACORE_VSS_4_i_bondpad" cell=xxxxx orientation=R180 )
>        . . . 
>       ( inst name="u_pad_ring/pad_VCON_VIN_i_bondpad" cell=xxxxxD60LA orientation=R180 )
>       ( inst name="u_pad_ring/pad_VSS_3_i_bondpad" cell=xxxxx         space=25           orientation=R180 )        <<<<<<<<<<<<
>       ( inst name="u_pad_ring/pad_DCORE_VDD_2_i_bondpad" cell=xxxxx      space=15        orientation=R180 )        <<<<<<<<<<<<
> ```
> 
> From the pads side, this can be done on each cheep, modifying the cheep.io.tpl. However, from the margins' side, this is hardcoded into the PadRing configuration, if I'm not wrong. It should be as simple as being able to split the margins per side

On the pads I allow explicit offset and skip declaration already (am writing unit tests to make sure), and the side offsets already wrote something that could work will test with unit tests aswell, will try to make it ready for tonight

JuanSapriza:
The only thing i would change about the pads logic, if it's not too much work, is to NOT set the space, offset, etc,  at the mcu-gen level, but rather give the freedom to have this set from the hjson or python. 

as an example, right now i am trying to set an non-uniform spacing for my pads (because i place physical cells in between), so i will need to specify "all these pads have a spacing of 15, but this one has a space of 25, and the following of 0". From the top-level python abstraction that's pretty straightforward in the current implementation (:wink: thanks), but these would be replaced by the automagically computed.

My suggestion (really, not necessarily for this PR), would be to simplify the process. Let the user, at the top level, define what are the spacings they want.

marinPh:
> The only thing i would change about the pads logic, if it's not too much work, is to NOT set the space, offset, etc, at the mcu-gen level, but rather give the freedom to have this set from the hjson or python.
> 
> as an example, right now i am trying to set an non-uniform spacing for my pads (because i place physical cells in between), so i will need to specify "all these pads have a spacing of 15, but this one has a space of 25, and the following of 0". From the top-level python abstraction that's pretty straightforward in the current implementation (:wink: thanks), but these would be replaced by the automagically computed.
> 
> My suggestion (really, not necessarily for this PR), would be to simplify the process. Let the user, at the top level, define what are the spacings they want.
> 
> Edit:
> 
> Maybe one easy way of overriding this is to NOT split things into PadDef and Pad (which i honestly still cannot understand why are different). Because i added a lot of custom attributes to my top level PadDef elements, but then those are lost in the conversion and I cannot use them on my templates

I see completly the point, I will look into spacing granularity. And for the pad padDef  "duplication" I agree, the point at the time was to modify as little the logic behind PadRing that is used for some logic I did not understand at the time

marinPh:
In the current documentation it is unclear if skip and spacing conventioned as the distance between the current and previous pad or the current the next pad. By reading the code I assume we are calculating from the distance between the previous and current and am using the same thing for bp_spacing

PR #828, opened 2025-12-03
Interrupt autogen

marinPh:
branch from pad_cfg.py that adds interrupt creation in .py


davidmallasen:
Missing documentation. Depends on #825

PR #832, opened 2025-12-05
Port to genesys2 dev board

danivz:
- Port X-HEEP to the Digilent Genesys2 development board.
- Update FPGA documentation and add Genesys2 use guide.

### DIFF
--- a/docs/source/FPGA/RunOnFPGA.md
+++ b/docs/source/FPGA/RunOnFPGA.md
@@ -13,8 +13,7 @@ The following FPGA boards are supported:
 | AMD         | [Zynq™ UltraScale+™ MPSoC ZCU104](https://www.amd.com/en/products/adaptive-socs-and-fpgas/evaluation-boards/zcu104.html)           | zcu104        |
 | AMD         | [Zynq™ UltraScale+™ MPSoC ZCU102](https://www.amd.com/en/products/adaptive-socs-and-fpgas/evaluation-boards/ek-u1-zcu102-g.html)   | zcu102        |
 | Digilent    | [Nexys-A7-100t](https://digilent.com/reference/programmable-logic/nexys-a7/start)                                                  | nexys-a7-100t |
-| Digilent    | [Genesys2](https://digilent.com/reference/programmable-logic/genesys-2/start)                                                      |
-genesys2      |
+| Digilent    | [Genesys2](https://digilent.com/reference/programmable-logic/genesys-2/start)                                                      | genesys2      |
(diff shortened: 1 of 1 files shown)


hw/fpga/xilinx_core_v_mini_mcu_wrapper.sv
```diff
@@ -99,6 +105,25 @@ module xilinx_core_v_mini_mcu_wrapper
...
+    end else begin
+      if (pwm_ctr + 1 == 10'd600) begin
+        pwm_ctr <= '0;
+      end else begin
+        pwm_ctr <= pwm_ctr + 1;
+      end
+    end
+  end
+
+  assign fan_pwm_o = pwm_ctr < 10'd75 ? 1'b1 : 1'b0;
```
davideschiavone: hi @danivz , can you explain why (in a comment) why you need to drive the fan from the logic? is it mandatory? if yes, can you point to the documentation? or if not, can you add a comment about why?

davideschiavone (changes requested):

sw/device/target/genesys2/x-heep.h
LuigiGiuffrida98, line 1: Hi @danivz can you please write the defines in this file like they are in sw/device/target/pynq-z2/x-heep.h?

LuigiGiuffrida98 (changes requested):

hw/fpga/xilinx_core_v_mini_mcu_wrapper.sv
danivz: I was driving it with a low-duty-cycle PWM to reduce the cooling fan’s noise. I wasn’t monitoring the FPGA temperature to adjust the fan speed accordingly, so I’m removing the fan-control logic. The fan will now run at full speed by default, in case someone wants to implement a high performance design consuming a lot of power.

sw/device/target/genesys2/x-heep.h
danivz: Done. I also updated the `x-heep.h` for ZCU102 board

docs/source/FPGA/Genesys2.md
davidmallasen, line 1: Should we add to the diagram where are the reset button and the execute from flash/on chip switches?

danivz: I've updated the block diagram. This information is missing in the other FPGA boards.

davidmallasen: Thank you Dani. I have it in mind it's missing for the other boards, I'm trying to make the docs more complete bit by bit, so we can already have it for the genesys II 😄

davideschiavone (approved):

PR #841, opened 2025-12-15
Master Slave Autogen

marinPh:
The goal is to remove all hardcoded references to master and slave buses


davidmallasen:
Might be revisited in the future, but we won't implement this for now

PR #852, opened 2026-01-07
Parallel simulating on CI

marinPh:
PR allows scalable APP tests by having 6 apps per jobs to run in parrallel. We also allow a other simulators to be built and simulated in parrallel. The new CI/CD pipeline goes:
- unit test peripherals and pads (as before)
- then we test compilation and in parallel builld the simulators
- we use the simulators built to run the apps (we at most 6 apps per job)

We pass from aournd 1h05 to aournd 23min

Multiple things could be done to help the CI/CD even more:
- we could generate a coverage report for the tests (on the apps or the rtl)
- have a power analysis of the apps thanks to verilator
- performance comparison between tests (on the apps)
- a way to use a cached simulator if the hw/file is unchanged (with an intelligent way to clean and keep)
- generation of waveform for failed tests (with a 1-2 day retention)
- if more tests using pytest framework better integration (for pretty coverage reports and flexibility)
- could test -march=native for compilation of verilator to maximie speed
- allow verilator to have multiple threads to simulate (som research needs to be done on how much)


.github/workflows/ci.yml
```diff
@@ -43,39 +43,60 @@ jobs:
...
+
+      - name: Generate test matrix
+        id: generate-matrix
+        run: |
+          python3 .github/scripts/generate_test_matrix.py > matrix.json
+          echo "matrix=$(cat matrix.json)" >> $GITHUB_OUTPUT
+
   compile-apps:
-    name: Compile All Apps
+    name: Compile Apps (Batch ${{ matrix.batch }})
```
davidmallasen: I wouldn't do this for the app compilation, since it already goes quite fast. Here I saw that it runs in ~3 minutes per batch (*6) instead of 7 minutes.

.github/workflows/ci.yml
```diff
@@ -43,39 +43,60 @@ jobs:
...
+      - name: Compile subset of apps with GCC and Clang
         shell: bash -l {0}
         run: |
           make mcu-gen X_HEEP_CFG=configs/ci.hjson && \
-          python3 test/test_apps/test_apps.py --compilers gcc,clang,gcc --compiler-paths /tools/riscv --compiler-prefixes riscv32-corev-,riscv32-unknown-,riscv32-unknown- --table --compile-only
+          python3 test/test_apps/test_apps.py --apps "${{ matrix.apps }}" --compilers gcc,clang,gcc --compiler-paths /tools/riscv --compiler-prefixes riscv32-corev-,riscv32-unknown-,riscv32-unknown- --table --compile-only
 
   simulate-apps:
-    name: Simulate All Apps
+    name: Simulate Apps (Batch ${{ matrix.batch }})
```
davidmallasen: This is repeating every time the docker pull, setup, mcu-gen, and verilator-build, which are quite time consuming. That's why the CI time goes down from 1h5min to 23ish minutes with 6 batches. Is there a way of not repeating all the setup?

davidmallasen (commented):
Any feedback/comments @StMiky @LuigiGiuffrida98 ?

LuigiGiuffrida98 (commented):
Looks good to me.

Would be nice to cache the ventilator model to avoid rebuilding (as @davidmallasen suggests).

Maybe, can be useful to have a job that only builds the model and cache the build folder to be reused by the other jobs. Pay attention to cleanup the cached files in case, it is important to avoid storing data on GitHub servers due to usage limits.

.github/workflows/ci.yml
marinPh: I am looking into it, but sharing verilator builds across jobs is harder than expected, artifacts don't work will try caches

test/test_apps/test_apps.py
```diff
@@ -174,60 +174,71 @@ def compile_app(an_app, compiler_path, compiler_prefix, compiler, linker, dry_ru
         return True
 
 
-def run_app(an_app, simulator, dry_run=False, verbose=True):
+def run_app(an_app, simulator, args, dry_run=False, verbose=True): # Note: added args param
```
LuigiGiuffrida98: Can you please remove this comment?

LuigiGiuffrida98 (approved):
Looks good to me, simply try to add the -j$(nproc) to the core file and then we are done from my side.

Makefile
davidmallasen, line 1: Can you remove this change?

.github/workflows/ci.yml
```diff
@@ -13,6 +13,13 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
+env:
+  # Artifact retention (can be used in steps context)
+  ARTIFACT_RETENTION_DAYS: 1
```
davidmallasen: This means that logs from the CI would also only be kept for 1 day instead of the usual (I think 90). I don't like this. I guess you added this to not keep many built simulators, right? Is there a way of improving this?

.github/workflows/ci.yml
```diff
@@ -43,39 +50,155 @@ jobs:
...
+        run: |
+          python3 .github/scripts/generate_test_matrix.py > matrix.json
+          echo "matrix=$(cat matrix.json)" >> $GITHUB_OUTPUT
+
+  # 1. BUILD SIMULATOR (HW)
+  build-simulator:
+    name: Build ${{ matrix.simulator }}
+    runs-on: ubuntu-latest
+    needs: [determine-image-tag, generate-app-matrix, lint, gen-peripherals, gen-pads, check-vendor, black-formatter]
+    timeout-minutes: 20  # Verilator build typically takes ~10 min
```
davidmallasen: Do we need a timeout here? If this is the case, why? I don't believe we have one now in the `build_simulator` function of `test_apps.py`

.github/workflows/ci.yml
```diff
@@ -43,39 +50,155 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-project
+      - name: Compile Only
         shell: bash -l {0}
         run: |
-          make mcu-gen X_HEEP_CFG=configs/ci.hjson && \
-          python3 test/test_apps/test_apps.py --compilers gcc,clang,gcc --compiler-paths /tools/riscv --compiler-prefixes riscv32-corev-,riscv32-unknown-,riscv32-unknown- --table --compile-only
+          make mcu-gen X_HEEP_CFG=configs/ci.hjson
+          # Compiles to sw/build/ directory
+          python3 test/test_apps/test_apps.py --compilers gcc --compile-only
```
davidmallasen: Why changing the parameters here? I don't believe this is correct

.github/workflows/ci.yml
```diff
@@ -43,39 +50,155 @@ jobs:
...
+      fail-fast: false
     container:
       image: ghcr.io/x-heep/x-heep/x-heep-toolchain:${{ needs.determine-image-tag.outputs.tag }}
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v4
-      - name: Setup Project
-        uses: ./.github/actions/setup-project
-      - name: Compile all apps with GCC and Clang
+      - uses: actions/checkout@v4
```
davidmallasen: Why remove/change the names here?

.github/workflows/ci.yml
```diff
@@ -43,39 +50,155 @@ jobs:
...
+          retention-days: ${{ env.ARTIFACT_RETENTION_DAYS }}
+          if-no-files-found: error
+
+  # 2. COMPILE APPS (SW)
   compile-apps:
-    name: Compile All Apps
+    name: Compile all Apps
     runs-on: ubuntu-latest
-    needs: [determine-image-tag, lint, gen-peripherals, gen-pads, check-vendor, black-formatter]
+    needs: [determine-image-tag, generate-app-matrix, lint, gen-peripherals, gen-pads, check-vendor, black-formatter]
```
davidmallasen: This job shouldn't need the `generate-app-matrix` anymore

.github/workflows/ci.yml
```diff
@@ -43,39 +50,155 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Generate test matrix
+        id: generate-matrix
+        run: |
+          python3 .github/scripts/generate_test_matrix.py > matrix.json
+          echo "matrix=$(cat matrix.json)" >> $GITHUB_OUTPUT
+
+  # 1. BUILD SIMULATOR (HW)
```
davidmallasen: I don't like these 1, 2, 3, comments. The order is not the case, as 2 is independent of 1. And this comment can be written as part of the name/description better.

.github/workflows/ci.yml
```diff
@@ -43,39 +50,155 @@ jobs:
...
+
 
+  # 3. RUN SIMULATION (Distributed)
   simulate-apps:
-    name: Simulate All Apps
+    name: Simulate (${{ matrix.simulator }}, Batch ${{ matrix.batch }})
     runs-on: ubuntu-latest
-    needs: [determine-image-tag, lint, gen-peripherals, gen-pads, check-vendor, black-formatter]
+    needs: [determine-image-tag, generate-app-matrix, build-simulator, compile-apps, lint, gen-peripherals, gen-pads, check-vendor, black-formatter]
+    timeout-minutes: 30  # 6 apps × ~3-5 min each + margin
```
davidmallasen: I don't think there is the need for timeout at the job level, this is done in the tests itself

.github/workflows/ci.yml
```diff
@@ -43,39 +50,155 @@ jobs:
...
     container:
       image: ghcr.io/x-heep/x-heep/x-heep-toolchain:${{ needs.determine-image-tag.outputs.tag }}
+
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v4
-      - name: Setup Project
-        uses: ./.github/actions/setup-project
-      - name: Compile and simulate all apps
+      - uses: actions/checkout@v4
```
davidmallasen: Again, why removing the names?

.github/workflows/ci.yml
```diff
@@ -43,39 +50,155 @@ jobs:
...
+
+      - name: Run Simulation
+        shell: bash -l {0}
+        run: |
+          # Regenerate MCU config (needed for headers/linker scripts)
+          make mcu-gen X_HEEP_CFG=configs/ci.hjson
+
+          # Compile and run simulations for this batch (simulator already built)
+          python3 test/test_apps/test_apps.py \
+            --apps "${{ matrix.apps }}" \
```
davidmallasen: Keep the parameters as before for `compiler-paths` and `compiler-prefixes`

.github/workflows/ci.yml
```diff
@@ -90,7 +213,7 @@ jobs:
         uses: ./.github/actions/setup-project
       - name: Mark Repository as Safe
         run: |
-          git config --global --add safe.directory $GITHUB_WORKSPACE
```
davidmallasen, line 93: Why?

test/test_apps/test_apps.py
```diff
@@ -174,60 +175,71 @@ def compile_app(an_app, compiler_path, compiler_prefix, compiler, linker, dry_ru
         return True
 
 
-def run_app(an_app, simulator, dry_run=False, verbose=True):
+def run_app(an_app, simulator, args, dry_run=False, verbose=True):
```
davidmallasen: Don't pass the full args, only what you need. Also, double-check the comment with the documentation of this function

test/test_apps/test_apps.py
```diff
@@ -174,60 +175,71 @@ def compile_app(an_app, compiler_path, compiler_prefix, compiler, linker, dry_ru
...
+        try:
+            # Run the binary directly
+            run_output = subprocess.run(
+                cmd,
+                capture_output=True,
+                timeout=SIM_TIMEOUT_S,
+                check=False 
+            )
+        except subprocess.TimeoutExpired:
+            print(BColors.FAIL + f"Timeout." + BColors.ENDC)
```
davidmallasen: Put a more clear string like was before, with the simulator and app name. Here and in all the other prints

test/test_apps/test_apps.py
```diff
@@ -174,60 +175,71 @@ def compile_app(an_app, compiler_path, compiler_prefix, compiler, linker, dry_ru
...
+                check=False,
             )
-            print(BColors.FAIL + str(run_output.stdout.decode("utf-8")) + BColors.ENDC)
-            return SimResult.FAILED
+        except subprocess.TimeoutExpired:
+            print(BColors.FAIL + f"Timeout." + BColors.ENDC)
+            return SimResult.TIMED_OUT
+
+    # --- COMMON RESULT PARSING ---
+    # This part remains exactly the same as your original script
```
davidmallasen: Thank you LLM, but remove this comment

test/test_apps/test_apps.py
```diff
@@ -174,60 +175,71 @@ def compile_app(an_app, compiler_path, compiler_prefix, compiler, linker, dry_ru
...
-            flush=True,
-        )
-    
-    if dry_run:
-        if verbose:
-            print(BColors.OKCYAN + f"[DRY RUN] make {simulator}-run" + BColors.ENDC, flush=True)
-        return SimResult.PASSED
+        print(BColors.OKBLUE + f"Running {an_app.name}..." + BColors.ENDC, flush=True)
+
+    if args.simulator_bin is not None and args.simulator_bin:
```
davidmallasen: Redundant

test/test_apps/test_apps.py
```diff
@@ -174,60 +175,71 @@ def compile_app(an_app, compiler_path, compiler_prefix, compiler, linker, dry_ru
...
+
+        hex_path = os.path.join(args.hex_dir, f"{an_app.name}.hex")
+        
+        if not os.path.exists(hex_path):
+             print(BColors.FAIL + f"Hex file not found: {hex_path}" + BColors.ENDC)
+             return SimResult.FAILED
+
+        cmd = ["make",f"{simulator}-run-app", an_app.name]
```
davidmallasen: Is this correct?

davidmallasen (changes requested):

Makefile
marinPh: The multi simulator build ?

davidmallasen: Here in the makefile I don't see any changes, just a whitespace deletion. Check the files changed of the PR

test/test_apps/test_apps.py
marinPh: reverted to old run_app what I tried failed

test/test_apps/test_apps.py
marinPh: reverted

test/test_apps/test_apps.py
marinPh: reverted couldn't make it work

.github/workflows/ci.yml
marinPh: it's to take in account the downloading aswell but can remove

.github/workflows/ci.yml
marinPh: Idk why but it did not let me use the downloaded simulator

davidmallasen:
Thanks guys. In the next commit we'll implement the `make_options` to speed up the verilator build (and also disable apps that time out), but for now we won't parallelize the simulation of apps in multiple jobs.

PR #853, opened 2026-01-07
Fix testbench

LuigiGiuffrida98:
When quadrilatero is not present in the testbench, the Questasim simulation fails due to an out of memory access.

The master ports of quadrilatero are not tied to 0 when it is not instantiated, so the X value (considered not 0 by modelsim) causes the systembus to get a request to the invalid address 0xXXXXXXXX.

This PR adds an else in the tb to fallback when quadrilatero is not present.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,7 +102,7 @@ MCU_GEN_TEMPLATES = \
 # Compiler options are 'gcc' (default) and 'clang'
 COMPILER 		?= gcc
 # Compiler prefix options are 'riscv32-corev-' (default) and 'riscv32-unknown-'
-COMPILER_PREFIX ?= riscv32-corev-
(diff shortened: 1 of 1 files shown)


hw/ip/boot_rom/Makefile
davidmallasen, line 1: This we were adding in the other PR. It needs the makefile adaptation I added yesterday so mcu-gen doesn't fail if not using the corev compiler

LuigiGiuffrida98: I think this should be merged asap, independently on the PR that brings this change 😆

davidmallasen: Sure, then add this too: https://github.com/x-heep/x-heep/pull/840/changes#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R105

davideschiavone (approved):

PR #860, opened 2026-01-13
:arrow_up: update common cells

davideschiavone:


hw/vendor/pulp_platform_common_cells.lock.hjson
```diff
@@ -9,6 +9,6 @@
   upstream:
   {
     url: https://github.com/pulp-platform/common_cells.git
-    rev: cf08cc1068e782638185b8316e1ebc0ef90596e7
+    rev: 9afda9abb565971649c2aa0985639c096f351171
```
LuigiGiuffrida98, line 12: Since this commit refers to a specific release of common cells, can't we just refer to a [version number](https://github.com/pulp-platform/common_cells/releases/tag/v1.38.0)? This way can be easier to check changelogs for future updates.

davideschiavone: makes sense

davideschiavone: @LuigiGiuffrida98 what is the version name?

annaburds: its [v1.39.0](https://github.com/pulp-platform/common_cells/releases/tag/v1.39.0)

LuigiGiuffrida98: no no, the commit in the vendor file is v1.38.0, while the one @annaburds is mentioning is the latest.

davideschiavone: we  need to use the same as of Chesire - @annaburds - can you check with @flague that we are all aligned? if not - I still prefer using the one of Chesire and then if you need a newer one, we can discuss separately

LuigiGiuffrida98: From what I see here https://github.com/pulp-platform/cheshire/blob/5752880791a2beabc75a5e39387ffdd2c3ce4230/Bender.yml#L23C3-L23C116 the version they use is v1.38.0

PR #868, opened 2026-01-23
Redo pad configuration

davidmallasen:
Attempt to redo the pad configuration in mcu-gen. Will replace #825.

Removes the mcu-gen cache since pickling was preventing the definition of external classes for the configuration. Being able to pass lists of files to mcu-gen removes the need for this caching. Now you can also pass directly a list of external mcu-gen templates so you can generate the MCU and the CHEEP in one mcu-gen.py call.

# ToDo list:

- [x] Check all `ToDo_padspy`
- [x] Port all X-HEEP templates to the new structure  @davidmallasen 
- [x] Vendorize on a CHEEP, port the configuration and the `cheep.io.tpl` 
- [x] Prepare an example "front-end" (`pad_cfg.py`) for X-HEEP @JuanSapriza 
- [x] Update documentation - @davidmallasen @JuanSapriza 

## To think: 
~❓ should `global_index` start on 0 or 1? Do physical pads also get a `global_index`?~
~❓ is there a need for `core_v_mini_mcu_pkg::side` being passed to the cell? is it ever used?~ 
~❓ What to print? what NOT to print?~
❓Any better solution for `spi(2)_sck_o` in `core_v_mini_mcu.sv.tpl` in HEEPokranios?
~❓Can we move the Pin and Pad subclasses from the pin.py and pad.py to the config?~

### DIFF
--- a/docs/source/Extending/eXtendingSW.md
+++ b/docs/source/Extending/eXtendingSW.md
@@ -216,6 +216,44 @@ export HEEP_DIR = <path_to_x_heep_relative_to_this_directory>
 ```
 </details><br>
 
+## Extending the mcu-gen configuration
+
+In your own `Makefile`, you can also extend the `mcu-gen` target to generate a custom MCU configuration for your application, includiing any `.tpl` files you might want to add. To do so, you can add an `mcu-gen` target in your `Makefile` like the following:
+
+<details>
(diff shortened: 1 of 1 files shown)


JuanSapriza:
Updated the templates based on what worked for heepokranios. 
hello world on verilator working

PR #879, opened 2026-02-05
`make venv` fails since latest pip update (26.0)

cousteaulecommandant:
There is a deprecated/invalid syntax issue in file `python-requirements.txt`:
```
# Development version with OT-specific changes
git+https://github.com/davideschiavone/fusesoc.git@ot#egg=fusesoc >= 1.11.0
```
That `>= 1.11.0` specifier is not valid for URLs, only for packages in the index.
This syntax has been deprecated since PIP 23.0 (from 2023), but until recently it merely triggered a warning.  However, since PIP 26.0 (released on 2026-01-31) the syntax is no longer supported at all and it causes an error when running `make venv`:[^1]
```
[...]
./.venv/bin/pip install -r util/python-requirements.txt -r docs/python-requirements.txt
error: invalid-egg-fragment

× The 'fusesoc >= 1.11.0' egg fragment is invalid
╰─> from 'git+https://github.com/davideschiavone/fusesoc.git@ot#egg=fusesoc >= 1.11.0'

hint: Version specifiers are silently ignored for URL references. Remove them. 
make: *** [Makefile.venv:233: .venv/bin/.initialized-with-Makefile.venv] Error 1
```
Note that, since `make venv` will always pull the *newest* version of PIP, this issue affects older commits retroactively, even those that used to work one week ago.

Removing the `>= 1.11.0` fixes the issue.  (Note that this doesn't do anything when used with git repos anyway.)

## Reproduction steps

1.  Be January 31 2026 or later.
2.  Get a machine with Python 3.9+ (3.8 will use PIP 25.0 at most which doesn't have this issue, since support for Python 3.8 was dropped on PIP 25.1).
3.  Clone a fresh copy of X-HEEP (or, if you already have an X-HEEP project, do `make clean-venv` or `rm -R .venv/` to ensure that there's no `.venv`).
4.  Run `make venv`.
5.  Observe the error.

[^1]: https://github.com/pypa/pip/commit/42154e8 "Remove support for non-bare egg fragments -- They were deprecated from all the way back to January 2023 🙃."

### DIFF
--- a/util/python-requirements.txt
+++ b/util/python-requirements.txt
@@ -8,10 +8,10 @@ black==24.8.0
 deepdiff
 
 # Development version with OT-specific changes
-git+https://github.com/davideschiavone/fusesoc.git@ot#egg=fusesoc >= 1.11.0
+git+https://github.com/x-heep/fusesoc.git@ot#egg=fusesoc
 
-# Install edalize by hand as pip install git+https://github.com/davideschiavone/edalize.git
-git+https://github.com/davideschiavone/edalize.git
+# Install edalize by hand as pip install git+https://github.com/x-heep/edalize.git
+git+https://github.com/x-heep/edalize.git
 
 # RISC-V firmware profiler
 git+https://github.com/vlsi-lab/rv_profile.git


davidmallasen:
Hello @cousteaulecommandant. As you say, this doesn't fail in Python 3.8, which is [the official version we support](https://x-heep.readthedocs.io/en/latest/GettingStarted/Setup.html#b-virtual-environment) (although I asked in [matrix](https://matrix.to/#/#x-heep:matrix.org) recently and there was consensus to update to 3.12 at some point soon). 

@davideschiavone  were those changes to edalize and fusesoc upstreamed and can we use the official repos there?

davideschiavone:
> Hello @cousteaulecommandant. As you say, this doesn't fail in Python 3.8, which is [the official version we support](https://x-heep.readthedocs.io/en/latest/GettingStarted/Setup.html#b-virtual-environment) (although I asked in [matrix](https://matrix.to/#/#x-heep:matrix.org) recently and there was consensus to update to 3.12 at some point soon).
> 
> @davideschiavone were those changes to edalize and fusesoc upstreamed and can we use the official repos there?

mmm I am not sure - I think I have a version which is in between upstream and openTitan... so better to rely on ours.. as you mention it - I moved the ESL forks to the x-heep github instead of my own :) 

@cousteaulecommandant - can you also update references to FuseSoc and/or Edalize from ESL or myown github to the X-HEEP ones?

https://github.com/x-heep/edalize

https://github.com/x-heep/fusesoc

cousteaulecommandant:
OK, done.  I kept the `@ot` part for fusesoc because I understand it was important, and also the `#egg=fusesoc` part (although that syntax seems to be deprecated - shall I remove it as well?)

davideschiavone:
> OK, done. I kept the `@ot` part for fusesoc because I understand it was important, and also the `#egg=fusesoc` part (although that syntax seems to be deprecated - shall I remove it as well?)

no idea to be honest.. if it works fine for me

cousteaulecommandant:
OK, then I guess the PR is good to go :) Pip doesn't warn about the `#egg=` syntax being deprecated so I suppose it'll do no harm to keep it just in case ("if it ain't broken...").

PR #882, opened 2026-02-11
🧱Move xheep_gen to different repository

LuigiGiuffrida98:
This PR is to move x_heep_gen to a new repository https://github.com/x-heep/xheep_gen.

Required actions for users when updating:
- Rename `x_heep_gen` to `xheep_gen` in your config files and templates. Here, paths like: `from x_heep_gen.xheep import XHeep` should now just be `from xheep import XHeep`.
- Any paths pointing to `util/x_heep_gen` should now be `util/mcu-gen/xheep_gen`.
- Any paths to `util/mcu_gen.py` should now be `util/xheep_gen/mcu_gen.py`.


davidmallasen:
This looks good to me, although like we mentioned in https://github.com/x-heep/x-alp/pull/24#discussion_r2996083212 , the idea would be to move mcu-gen.py as well into xheep_gen. In that case, I think it'd be better to do it as part of this PR as otherwise from the user perspective we might have to change paths twice in the CHEEPs

davidmallasen (approved):

PR #886, opened 2026-02-13
FPGA support SL

annaburds:


sw/applications/example_serial_link_emulation_receive/main.c
```diff
@@ -0,0 +1,47 @@
+// Copyright 2025 EPFL
```
davideschiavone: @annaburds - please use the same application that you used for simulation, just add ifdef for the FPGA to avoid having too many applications - see https://github.com/x-heep/x-heep/blob/main/sw/applications/demo_spi_master_slave/main.c

annaburds:
That is **not** the final version. Most of the files to be adjusted and cleaned.

hw/fpga/constraints/pynq-z2/pin_assign.xdc.tpl
davidmallasen, line 1: This has changed in the recently merged https://github.com/x-heep/x-heep/pull/914

configs/pad_cfg.py
```diff
@@ -64,6 +64,16 @@ def config() -> PadRing:
         Inout("spi2_sd_3"),
         Inout("i2c_scl"),
         Inout("i2c_sda"),
+        Input("ddr_rcv_clk_i"),
```
davideschiavone: @davidmallasen , if the PADs (see below) are muxed with GPIOs, and GPIOs are the main ones, why do we need these PADs here?

davidmallasen: These here are not pads, they are pins, aka rtl signals that then automagically create the pads. For details you can check the overview here: https://x-heep.readthedocs.io/en/latest/Configuration/PadConfiguration.html#overview

hw/fpga/constraints/pynq-z2/pin_assign.xdc.tpl
```diff
@@ -0,0 +1,114 @@
...
+set_property -dict {PACKAGE_PIN V13 IOSTANDARD LVCMOS33} [get_ports {spi_slave_mosi_io}] ; # arduino_gpio_tri_io[3]
+
+# PDM2PCM
+set_property -dict {PACKAGE_PIN Y9 IOSTANDARD LVCMOS33} [get_ports {pdm2pcm_clk_io}] ; # rpi_gpio_tri_io[13]
+set_property -dict {PACKAGE_PIN A20 IOSTANDARD LVCMOS33} [get_ports {pdm2pcm_pdm_io}] ; # rpi_gpio_tri_io[12]
+
+# I2S
+set_property -dict {PACKAGE_PIN B19 IOSTANDARD LVCMOS33} [get_ports {i2s_sck_io}] ; # rpi_gpio_tri_io[8]
+set_property -dict {PACKAGE_PIN B20 IOSTANDARD LVCMOS33} [get_ports {i2s_ws_io}] ; # rpi_gpio_tri_io[4]
+set_property -dict {PACKAGE_PIN P15 IOSTANDARD LVCMOS33} [get_ports {i2s_sd_io}] ; # arduino_direct_iic_scl_io
```
davidmallasen: I believe you missed this line from https://github.com/x-heep/x-heep/pull/914

hw/fpga/xilinx_core_v_mini_mcu_wrapper.sv.tpl
LuigiGiuffrida98, line 1: Why is this file a template? Seems that the user_peripheral_domain variable is never used in the file.

davidmallasen: You're totally right. @annaburds ?

annaburds: Yep, after I changed the pads I removed from here the logic. Will delete it. Thanks

hw/core-v-mini-mcu/core_v_mini_mcu.sv
```diff
@@ -420,6 +430,54 @@ module core_v_mini_mcu
...
+  //isogate exposed outside for UPF sim flow and switch cells
+  assign memory_subsystem_banks_powergate_iso_n[8] = memory_subsystem_pwr_ctrl_out[8].isogate_en_n;
+  assign memory_subsystem_banks_set_retentive_n[8] = memory_subsystem_pwr_ctrl_out[8].retentive_en_n;
+  assign memory_subsystem_clkgate_en_n[8] = memory_subsystem_pwr_ctrl_out[8].clkgate_en_n;
+  assign memory_subsystem_banks_powergate_switch_n[9] = memory_subsystem_pwr_ctrl_out[9].pwrgate_en_n;
+  assign memory_subsystem_pwr_ctrl_in[9].pwrgate_ack_n = memory_subsystem_banks_powergate_switch_ack_n[9];
+  //isogate exposed outside for UPF sim flow and switch cells
+  assign memory_subsystem_banks_powergate_iso_n[9] = memory_subsystem_pwr_ctrl_out[9].isogate_en_n;
+  assign memory_subsystem_banks_set_retentive_n[9] = memory_subsystem_pwr_ctrl_out[9].retentive_en_n;
+  assign memory_subsystem_clkgate_en_n[9] = memory_subsystem_pwr_ctrl_out[9].clkgate_en_n;
```
LuigiGiuffrida98: Is this the output of the default X-HEEP configuration?

annaburds: there is a bug w mcu-gen. it was using more RAMs then should. CI cant catch it. 
Just pushed a newer version, which should work w 2 when mcu-gen is fixed

PR #889, opened 2026-02-19
MacOS install scripts

Ludblanc:
Here I propose 2 scripts:

`util/install_tools_and_conda_env.zsh` allow to install the tools with specific versions and fix some compatibility issues such as pyYAML, and compilation issues with the specific version of verilator required for x-heep.

`util/start_env.zsh` is a script to source to start the environment.

The two scripts works on both macOS arm64 and intel, as well as intel ubuntu.

Incidentally the python version had to be updated for compatibility issues.

I tried to update the documentation accordingly


LuigiGiuffrida98:
I completely uninstalled conda (removing the environment too), verilator and the toolchain to restart from scratch. I reinstalled conda and run your script.

hjson and some other packages are installed as a dependency of fusesoc/edalize. But pylibfst installation fails because of cmake versioning errors. The workaround is to use a specific version of python instead of letting conda inherit it from the system (python 3.8 should be fine) python >= 3.10 deprecates the support for the version of cmake required by the version of pylibfst installed by edalize.

(This is the issues I encountered when installing the conda env on my mac and the workaround I used to fix things).

I will try it asap and push the fix to this branch

LuigiGiuffrida98 (approved):

util/conda_environment.yml
```diff
@@ -2,8 +2,10 @@ name: core-v-mini-mcu
 channels:
   - conda-forge
 dependencies:
-  - python=3.8
+  - python=3.11
```
davidmallasen, line 5: We are planning to update to python 3.12 as discussed in #891 and #879, so I would keep that update separate from here. This includes the update setuptools that you also added here

docs/source/GettingStarted/Setup.md
```diff
@@ -54,6 +55,22 @@ libfl2 libfl-dev zlibc zlib1g zlib1g-dev
 
 Links for the packages relative to each software can also be found under the corresponding section of this guide. In general, make sure to have a look at the [Check system requirements](https://opentitan.org/book/doc/getting_started/index.html) section of the OpenTitan documentation.
 
+The script `util/install_tools_and_conda_env.sh` can be used to install the required dependencies and set up the Conda environment on Linux. It uses `apt` to install the required packages and then sets up a Conda environment with the necessary Python dependencies. It also installs the CORE-V toolchain, Verilator and the oss-cad-suite. You can select the versions of the tools you want to install by changing the corresponding variables at the beginning of the script. After running the script, you can activate the environment with `source util/start_env.sh`.
```
davidmallasen, line 58: Where are these `.sh` versions? I only see `.zsh`.

util/install_tools_and_conda_env.zsh
```diff
@@ -0,0 +1,351 @@
...
+        error "${label} failed:"
+        echo "$output"
+        exit 1
+    fi
+}
+
+# --- Configuration & Versions ---
+RISCV_V_DATE="20240530"
+OSS_CAD_DATE="2026-02-15"
+VERIBLE_HASH="v0.0-4051-g9fdb4057"
```
davidmallasen, line 37: This is not the version we have in the documentation. It should be `v0.0-4023-gc1271a00`

util/conda_environment.yml
```diff
@@ -2,8 +2,10 @@ name: core-v-mini-mcu
...
   - pip
+  - pyYAML
```
davidmallasen, line 7: We should add a version if we move it from the requirements to here

docs/source/GettingStarted/Setup.md
davidmallasen, line 58: Clarify that this is only a shortcut that should be understood rather that run blindly. For both this and macos

davidmallasen (changes requested):
@davideschiavone @StMiky what do you think of adding these scripts to the repo? I'm not entirely convinced as could be an unreliable shortcut in different systems, but on the other hand could be handy

LuigiGiuffrida98:
Hi @Ludblanc what is the status of this PR?

Ludblanc:
Hi @LuigiGiuffrida98,
Just for context: I've been using x-heep as a hobby project for 2 weeks in March, not as part of my PhD. I'll have more time to contribute after my next tapeout.
The motivation for this PR is that I installed it this way on my machine while supervising a student who is trying to compile x-heep for Lattice FPGA, and I thought it could be useful for other people using Macs.
A couple of notes based on @davidmallasen's feedback: the Python version was picked because with 3.8 the conda env couldn't be created on macOS, and the config targets zsh since it's the default shell on Mac.
If anything isn't working or needs small adjustments, I'm happy to take care of it.

PR #894, opened 2026-02-23
move dma to vendor and fix verilator 4

davideschiavone:


davideschiavone:
@davidmallasen , @LuigiGiuffrida98 ,  @StMiky  when executing

```
find . -name '*.vendor.hjson' | xargs -n1 util/vendor.py --verbose &&   util/git-diff.py --error-msg '::error ::Found differences, please re-vendor.'
```

I do not see any diff in my terminal, but the CI fails - do you know why?

PR #903, opened 2026-03-04
Add questasim-run in makefile

Allain18:
- Add `make questasim-run ` of the same style as `make verilator-run`. 
- Update doc

### DIFF
--- a/docs/source/How_to/Simulate.md
+++ b/docs/source/How_to/Simulate.md
@@ -135,7 +135,7 @@ make questasim-build-opt
 Then, you can run the compiled application with:
 
 ```bash
-make questasim-build-opt-run
+make questasim-run-opt
 ```
 
 Alternatively, go to


Makefile
```diff
@@ -231,9 +231,7 @@ verilator-run-helloworld: mcu-gen verilator-build
 		--run_options="+firmware=../../../sw/build/main.hex $(SIM_ARGS)"
 
 ## First builds the app and then uses Verilator to simulate the HW model and run the FW
-verilator-run-app: app
```
davidmallasen, line 234: Please undo this change. The dependencies of make targets are not guaranteed to be executed in the order that you specify them. Same in the other `run-app` flavors below

Allain18: Done

docs/source/How_to/Simulate.md
```diff
@@ -114,13 +120,25 @@ and type to run your compiled software:
...
 
 ```bash
 make questasim-build-opt
 ```
 
-then go to
+Then, you can run the compiled application with:
+
+```bash
+make questasim-build-opt-run
```
davidmallasen: I think this does not correspond to the makefile, can you double-check?

Allain18: You're right, I mixed both commands

PR #904, opened 2026-03-09
:white_check_mark: fix power manager hal and app

davideschiavone:


sw/linker/link.ld.tpl
```diff
@@ -234,11 +234,12 @@ SECTIONS
   _edata = .; PROVIDE (edata = .);
   . = .;
 
-  .power_manager : ALIGN(4096)
-  {
-     PROVIDE(__power_manager_start = .);
-     . += 256;
-  } >ram1
+    .power_manager : ALIGN_WITH_INPUT
```
davideschiavone: @AlessandroVaraldi - I changed it so that it has the same alignment constraints as the FLASH_LOAD one... can you check it works for you?

AlessandroVaraldi: I checked it and it works for me, thanks

davideschiavone: sorry @AlessandroVaraldi I noticed I push the wrong version, did you try the newest?

AlessandroVaraldi: I tried the new version too and it works for me

PR #908, opened 2026-03-12
Casted all mux signals into the same length

JuanSapriza:
Because the muxing is done in a matrix, all registers need to be of the same size so we can simply assign them

```sv    
output logic [NUM_PAD-1:0][2:0] pad_muxes_o
```


JuanSapriza:
image

Now it will look like this

PR #912, opened 2026-03-13
refactor dma and waiver for v5

davideschiavone:
waits for https://github.com/x-heep/x-heep/pull/906 and also update vendor to remove path constraints


hw/core-v-mini-mcu/include/x-heep_structs.core
```diff
@@ -0,0 +1,22 @@
+CAPI=2:
+
+name: "x-heep::structs"
```
StMiky: `x-heep:rtl:structs`

StMiky (changes requested):

PR #913, opened 2026-03-13
Open-Source Synthesis with Yosys

marcop0022:
I started performing an open-source flow to synthesize with Yosys and the Slang plugin.


davideschiavone:
@marcop0022 , @LuigiGiuffrida98 , it is not a good idea to vendorize the PDK... too big

PR #854, opened 2026-01-08
:construction: :zap: :fire: add flash controller (#795)

davideschiavone:
Found 3 bugs unrelated to the Flash Controller but to be fixed them here before merge:

1. DMA reads one extra address
2. DMA does not respect OBI (retract REQ) if SLOT && REQ && !GNT and next cycle  !SLOT
3. SLOTs from SPI misscommunication (cycle-wise) with DMA

Before merging:

- [x] clean up HW
- [x] disable controller by default except for CI hjson
- [x] move spi host to its own repository under xheep
- [x] add documentation and tell that this is not tested on FPGA
- [x] fix DMA OBI protocol
- [x] fix extra READ request
- [x] fix SLOT communication by waiting rvalid
- [x] add to fixed SLOT communication to wait for rvalid AND software counter
- [x] test FPGA without the flash_controller (at least read/write from FLASH quad and single mode) (@danivz @LuigiGiuffrida98 )
- [ ] check combinatorial loops
- [x] check HW slots @alessionaclerio22 
- [x] check flash_load

PR #916, opened 2026-03-16
🔧 fix SPI FLASH write on the FPGA

Patrick-Pataky:
The controller implemented in #854 didn't work out of the box on FPGA (pynq-z2), on the write path. This PR fixes those bugs and we can now write on multiple pages/sectors.


hw/ip/w25q128jw_controller/rtl/w25q128jw_controller.sv
```diff
@@ -712,25 +547,22 @@ module w25q128jw_controller
...
           // We need to preserve other bits when modifying RXWM (preserved in read_value from OBI FSM)
           FWAIT_SET_RXWM_R: begin
-            spi_host_reg_req_offset  = SPI_HOST_COMMAND_OFFSET;
+            spi_host_reg_req_offset  = SPI_HOST_CONTROL_OFFSET;
             spi_host_reg_req_o.write = 1'b0;
             spi_host_reg_req_o.valid = 1'b1;
             if (spi_host_reg_rsp_i.ready && ~spi_host_reg_rsp_i.error) begin
-              //we are sharing the sector_iter_offset_q register with the rdata from SPI to save resources
-              sector_iter_offset_d = spi_host_reg_rsp_i.rdata;
+              spi_control_d = spi_host_reg_rsp_i.rdata;
```
davideschiavone, line 667: is there a reason why we cannot use the sector_iter_offset_d register? I know it's not its right name, but we save one extra register at least :)

sw/device/bsp/w25q/w25q.c
```diff
@@ -265,7 +265,7 @@ w25q_error_codes_t w25q128jw_init(spi_host_t* spi_host) {
     // Power up flash
     flash_power_up();
 
-    // Set QE bit (only FPGA, simulation do not support status registers at all)
+    // Set QE bit (only FPGA, simulation do not support writing to status registers)
     #ifndef TARGET_SIM
     if (set_QE_bit() == FLASH_ERROR) return FLASH_ERROR; // Error occurred while setting QE bit
```
davideschiavone, line 269: it probably makes sense to add this feature, in this or next PR

hw/ip/w25q128jw_controller/rtl/w25q128jw_controller.sv
```diff
@@ -357,6 +346,13 @@ module w25q128jw_controller
       // Wait for SW to set the START bit in CONTROL register
       TOP_IDLE: begin
         if (reg2hw.control.start.q) begin
+          if (!reg2hw.control.rnw.q) begin
```
davideschiavone: can you explain me this part? @Patrick-Pataky

hw/ip/w25q128jw_controller/rtl/w25q128jw_controller.sv
Patrick-Pataky: Yes, during a write on multiple sectors, we increment this register (see line 1283):
```
sector_iter_offset_d = sector_iter_offset_q + {19'b0, SE_BSIZE};
```
Since for each page write (at thus each sector write) we need to go through FWAIT, this register gets overwritten, causing an infinite loop in simulation.

sw/device/bsp/w25q/w25q.c
Patrick-Pataky: Will be implemented in this PR!

hw/ip/w25q128jw_controller/rtl/w25q128jw_controller.sv
Patrick-Pataky: It might be not needed, it's more a safeguard to make sure that every counter is reset at the beginning of every operation, to ensure a clean continuation.
Do you want me to remove it, and make sure it works without?

hw/ip/w25q128jw_controller/rtl/w25q128jw_controller.sv
davideschiavone: makes sense thx

hw/ip/w25q128jw_controller/rtl/w25q128jw_controller.sv
davideschiavone: yes I think it is good to let the SW do that, or an explicit state at the end of READ and or WRITE which reset every state before going back to IDLE - but I think the SW approach is preferred to me

Patrick-Pataky: Sounds good, I'll test it now with that modification before pushing.

PR #924, opened 2026-03-26
[flow] Check patch consistency on Vendor dependencies

StMiky:
Update the `util/check-vendor.py` script to include patches in the consistency check run after `vendor.py`. This check builds a list of patches and enforces that the name and MD5 hash of the applied patches matches in every project importing the IP. This prevents subtle and opaque errors from being introduced when the same dependency appears multiple times in a parent project, where only one copy of such dependency is kept.


hw/vendor/openhwgroup/cv32e40x/rtl/cv32e40x_cs_registers.sv
```diff
@@ -93,7 +93,7 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
   // Interrupts
   input  logic [31:0]                   mip_i,
   input  logic                          mnxti_irq_pending_i,
-  input  logic [CLIC_ID_WIDTH-1:0]    mnxti_irq_id_i,
```
davideschiavone, line 96: @StMiky is this necessary?

hw/vendor/pulp_platform/axi/src/axi_burst_splitter_gran.sv
```diff
@@ -63,7 +63,7 @@ module axi_burst_splitter_gran #(
   axi_resp_t  slv_resp, act_resp, unsupported_resp;
 
   axi_multicut #(
-    .NoCuts    ( CutPath  ),
+    .NoCuts    ( 32'(CutPath)  ),
```
davideschiavone: why? @StMiky

hw/vendor/openhwgroup/cv32e40x/rtl/cv32e40x_cs_registers.sv
StMiky: No, I removed it

hw/vendor/pulp_platform/axi/src/axi_burst_splitter_gran.sv
StMiky: Width mismatch. It's waivable so I removed it as well.

PR #928, opened 2026-03-27
✂️ separate .cores for sim files

davideschiavone:


hw/simulation/pad_sim.core
```diff
@@ -0,0 +1,31 @@
...
+
+  files_verilator_waiver:
+    files:
+    - pad_sim.vlt
+    file_type: vlt
+
+targets:
+  default:
+    filesets:
+    - target_sim ? (tool_verilator? (files_verilator_waiver))
```
davidmallasen: These extra `target_sim ?` are to avoid mistakes I guess, right? In theory that kind of check should be done at the top-level `.core` in case your simulation target in the CHEEPs is not named `sim` or `sim_sc`

hw/simulation/sram_sim.core
```diff
@@ -0,0 +1,26 @@
...
+# SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+
+
+filesets:
+  files_sim:
+    files:
+    - sram_wrapper.sv
+    file_type: systemVerilogSource
+
+  files_verilator_waiver:
```
davidmallasen, line 17: Not used

hw/simulation/pad_sim.core
davideschiavone: yes are done for this reason - I can move them to the top level .core if you prefer

hw/simulation/sram_sim.core
davideschiavone: what is not used?

hw/simulation/pad_sim.core
davidmallasen: I think it would be better, yes. Otherwise we might need to modify those for future targets or add patches that can end up messy

core-v-mini-mcu.core
```diff
@@ -60,14 +60,8 @@ filesets:
...
     - pulp-platform.org::tech_cells_generic
-    files:
-    - hw/simulation/sram_wrapper.sv
-    - hw/simulation/pad_cell_bypass_input.sv
-    - hw/simulation/pad_cell_bypass_output.sv
-    - hw/simulation/pad_cell_inout.sv
-    - hw/simulation/pad_cell_input.sv
-    - hw/simulation/pad_cell_output.sv
-    file_type: systemVerilogSource
+    - x-heep:sim:sram_wrapper
```
davideschiavone, line 63: it is used here @davidmallasen

hw/simulation/sram_sim.core
davidmallasen: The `files_verilator_waiver` fileset is not used in the default target, so not exposed

davidmallasen: Missing something like `tool_verilator? (files_verilator_waiver)`

hw/simulation/pad_sim.core
davideschiavone: done

hw/simulation/sram_sim.core
davideschiavone: you r right, done

PR #941, opened 2026-03-31
move read-only sections to ram0

AlessandroVaraldi:
Align link.ld.tpl and flash linker templates by separating .rodata from .text and placing it in ram0 (code-side). This keeps read-only data consistently on the code bank across on-chip, flash_load, and flash_exec layouts. Update mem_usage.py accordingly: classify .rodata separately and adjust the report to show a bank‑level breakdown (ram0/ram1/etc.) with a clear contents column, which is a more accurate representation of how memory is partitioned than the previous region breakdown

### DIFF
--- a/scripts/building/mem_usage.py
+++ b/scripts/building/mem_usage.py
@@ -367,7 +367,10 @@ def get_regions(section_headers):
         elif section_name in flash_data_sections:
             region_type = "f"
             name = "FLASH data"
-        elif section_name in code_sections or "X" in section["flags"] or "W" not in section["flags"]:
+        elif "W" not in section["flags"] and "X" not in section["flags"]:
+            region_type = "R"
+            name = "rodata"
+        elif section_name in code_sections or "X" in section["flags"]:
             region_type = "C"
             name = "code"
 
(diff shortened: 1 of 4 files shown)


scripts/building/mem_usage.py
```diff
@@ -367,7 +367,7 @@ def get_regions(section_headers):
         elif section_name in flash_data_sections:
             region_type = "f"
             name = "FLASH data"
-        elif section_name in code_sections or "X" in section["flags"]:
+        elif section_name in code_sections or "X" in section["flags"] or "W" not in section["flags"]:
```
JuanSapriza: I agree with Jose that it would be nice to actually distinguish ro data from code. Could we instead add another elif and give it region_type = 'R'?

JuanSapriza: @AlessandroVaraldi  sorry for the delay btw

AlessandroVaraldi: Hi @JuanSapriza, I agree that read-only data should be separated from the code, but this would create an inconsistency with flash_load, where .rodata is included in the .text region. So we can either include .rodata in .text in on_chip as well, or separate it in flash_load. Which do you think is better?

JuanSapriza: aaaahhaa... i see. I dont know if knowing that a piece of memory is rodata is useful or not, so I don't mind any of the two. All ia was afraid was of seeing `CCCC--------CC` ant not knowing what that CC at the end was. But any solution is fine for me

AlessandroVaraldi: So I separated rodata from .text in all linker scripts, for consistency. Also I had to update mem_usage.py to show a breakdown of the banks, and not the memory regions (I think this way is more clear)

JuanSapriza: Agree!

PR #949, opened 2026-04-07
questasim workaround sl

annaburds:
Added simulation-only handling for GPIO-multiplexed serial link signals to avoid X/Z propagation in Questa, while preserving the original FPGA behavior.


davidmallasen:
These pads are GPIOs by default, so that's why they're inout from the start. The switch to be muxed to the other ddr signals has to happen later. I think the root of the problem must be somewhere else

PR #952, opened 2026-04-14
Rewriting READ/WRITE tests for the FLASH controller

Patrick-Pataky:
Added more extensive tests to test more cases, in a more regular way.
The library is also updated to be easier to use.

Format of the tests:

- Read:
```python
for each speed in [standard speed, quad speed] (with DMA):
   for each mode in [polling, interrupt]:
     test read with:
       - single sector
       - two sectors
       - single byte
       - unaligned single sector
       - unaligned cross sector
```

- Write:
The write tests are very similar, with the addition of testing an overlapping write over the two sectors right after doing the first one. This allows to verify that the FSM resets for every write.


sw/applications/example_w25q128jw_read/main.c
```diff
@@ -4,26 +4,41 @@
...
 #include "core_v_mini_mcu.h"
 #include "x-heep.h"
 #include "w25q128jw.h"
 
 #include "w25q128jw_controller.h"
-#include "sram_data.h"
-#include "csr.h" // For CSR macros
-#include "rv_plic.h" // For PLIC functions
+#include "data.h"
 #include "w25q128jw.h"
```
Allain18, line 20: Include 2 times "w25q128jw.h"

PR #953, opened 2026-04-14
Update vendor.py to apply global patches unconditionally

StMiky:
## Background
Currently, `vendor.py` applies patches inside the `patch_dir` of a `.vendor.hjson` file _only if_ there is no `mapping:` section in it. If a mapping is defined, then patches are applied only if a `patch_dir` field is defined for each mapping entry. The `patch_dir` specific of each mapping entry is defined relative to the global `patch_dir`. All the patches are applied _after_ copying the source files inside the target directory, so patches must be defined relative to the _target_ directory, that is `target_dir/to` in the case of a mapping entry. As a consequence, if a `mapping:` is defined, global patches are _silently_ ignored.

### Example
Consider the following example:
```yaml
{
  name: "lowrisc_opentitan",
  target_dir: "lowrisc_opentitan"

  upstream: {
    url: "https://github.com/lowRISC/opentitan.git",
    rev: "47a0f4798febd9e53dd131ef8c8c2b0255d8c139",
  },

  patch_dir: "patches/lowrisc_opentitan",

  mapping: [
    {from: "hw/ip/prim",          to: "hw/ip/prim",         patch_dir: "prim"},
    {from: "hw/ip/prim_generic",  to: "hw/ip/prim_generic", patch_dir: "prim_generic"},
  ]
```

In this case, `vendor.py` does the following:
1. Clone the upstream repo in `/tmp/whatever` and checkout to the specified `rev`.
2. For each entry in `mapping:`, copy the content of `/tmp/whatever/entry.from` into the target directory `target_dir/entry.to`.
3. For each patch in `patch_dir/entry.patch_dir`, apply the patch in `target_dir/entry.to`

Any patch inside the `patch_dir` directory is never applied.

## The Issue
The above approach ignores "global" patches (i.e., `.patch` files inside `patch_dir`) whenever the `.vendor.hjson` file contains a `mapping` field. Let's consider the following situation:
- X-HEEP vendorizes repo `A` in `hw/vendor/A` with patches inside `hw/vendor/patches/A`.
- Johhny wants to build a chip where X-HEEP is extended with his supercool accelerator `B`, that also vondorises `A`. However, `B` only needs a couple of files from `A`, so Johnny defines a `mapping:` field inside his `B/hw/vendor/A.vendor.hjson` file:
```yaml
    patch_dir: "patches/A",
    mapping: [
        {from: "file0.sv", to: "file0.sv"},
        {from: "file1.sv", to: "file1.sv"},
    ]
```
- To be sure that his accelerator still works when integrated in X-HEEP, Johnny make sure to put in his `A.vendor.hjson` file the same revision as X-HEEP's `A.vendor.hjson`, and also copies X-HEEP's patches for A inside his `B/hw/vendor/patches/A` directory.
- In the chip, Johnny imports his accelerator excluding the `hw/vendor/A` directory, as it is already available inside `hw/vendor/x-heep/hw/vendor/A`.

When Johnny runs
```bash
python3 util/vendor.py -vU hw/vendor/A.vendor.hjson
```
in his repository, **patches inside `B/hw/vendor/patches/A` are NOT applied**. Instead, `vendor.py` rejects the `.vendor.hjson` files as a global `patch_dir` is defined without any `mapping` entry containing local `patch_dir` fields.

If Johnny removes the patches, `vendor.py` suceeds, but he will test his accelerator with files from `A` that differ from their version in the chip, potentially resulting in broken silicon. Even worse, if he defines the `entry.patch_dir` fields but leaves the corresponding directories entries, the `vendor.py` succeeds but the global patches are silently ignored.

The only safe solution would be not using the `mapping` feature, but this means importing the whole `A` repository or writing an infinite `exclude_from_upstream` list, which is suboptimal.

## Proposed Solution
This PR modifies `vendor.py` as follows:
1. Make `vendor.py` apply global patches defined in `patch_dir` _unconditionally_ after cloning the repo to `/tmp/whatever` (i.e., before copying the files to `target_dir` and regardless of a `mapping:` field being defined).
2. Remove the requirement that rejects a global `patch_dir` from being defined if no `patch_dir` is defined in the `mapping` entries.

Therefore, global patches are always applied, and local patches are applied _after_ that, relative to the `entry.to` directory.

This solution allows projects with common dependencies to apply the same patches regardless of them using the `mapping` feature of Vendor, and most importantly prevents the possibility of having different dependencies despite specifying the same revision and patches in a `.vendor.hjson` file.

## Other changes
The PR also includes some minor changes that improve X-HEEP compatibility when vendorised in other projects, mainly X-ALP. Namely:
- Make the `--recursive` option in `vendor.py` optional (enable with the `upstream.clone_recursive = true` option in `*.vendor.hjson`)
- Add more files to `pulp_platform_register_interface.core` and add necessary waivers
- Add missing test files to `riscv_dbg` so the dependency is more self-contained for testing in parent projects


LuigiGiuffrida98 (approved):

util/vendor.py
davidmallasen: Update this comment with the new situation. I think it is:
- Remove the "and there is no mapping"
- Change "if and only if" to "if"

StMiky: Changed, thanks

PR #962, opened 2026-05-11
Allow pads to depend from X-HEEP config

StMiky:
This PR changes the signature of the `config()` function that defines the pads to accept a mandatory `XHeep` object as an argument:
```python
def config(xheep: XHeep) -> PadRing:
```

This enables using the `mcu-gen` configuration to parameterize the pad ring generation. For example, I2S pads might only be defined if the I2S peripheral is actually added to X-HEEP.


davidmallasen (approved):
Modify docs as discussed and LGTM

StMiky:
Docs updated with the new signature and a brief description

PR #969, opened 2026-05-27
Mem bridge for w25q128jw

### LINKED ISSUE #873: add memory mapped bridge to w25q128jw controller
The w25q128jw controller gets commands only explicitly from software - but we can replace and remove the Yosys XIP unit and use this controller to leverage the OpenTitan SPI Host and DMA to both READ and WRITE to the FLASH implicitly by means of lw/sw operations.

Right now, the Yosys IP only supports read operations. In addition, X-HEEP requires two SPIs, while by solving this issue, it would result in only one inside the SPI Flash Subsystem.

Allain18:
- Direct memory access work in read and write both on verilator (standard speed) and questasim (quad speed) 

- We can read and write byte short and word

- Unfortunately, for writing the bus NEEDS to be NtoM

- execute_from_flash is removed along with obi_spimemio, however "execute_from_flash" is still often in the docs.

- [ ] PR in [xheep_spi](https://github.com/x-heep/xheep_spi) created
- [ ] PR in [xheep_spi](https://github.com/x-heep/xheep_spi) merged

Close #873


davideschiavone:
this PR has been moved to #980 - thanks @Allain18

PR #970, opened 2026-06-02
Adding a simple write-back cache to the W25Q128JW flash controller  

Patrick-Pataky:
This PR adds a simple direct-mapped write-back cache to the W25Q128JW flash controller. The cache is meant to be highly parametrizable, with the following parameters:

| Parameter | Default Value | Description |
| :--- | :---: | :--- |
| `N_WAYS` | `1` | Number of cache ways. Currently only supports `1` (direct-mapped). |
| `N_SETS` | `4` | Number of sets (lines) in the cache. |
| `ADDR_WIDTH` | `24` | Width of the flash byte-address. A 24-bit width natively spans the entire 16\,MiB byte-addressable memory space of the flash chip. |
| `SECTOR_SIZE_BYTES` | `4096 (4KiB)` | Number of bytes in a flash sector. |

---

### Cache Line State & Address Mapping

Each cache set (line) is fixed to a full flash sector (`SECTOR_SIZE_BYTES` bytes) and has its own status using the following metadata state bits:

| State Bit | Width | Description |
| :--- | :---: | :--- |
| `Valid` | 1 bit | Indicates whether the cache line contains valid data. |
| `Dirty` | 1 bit | Indicates whether the cached data has been modified by a write operation and is not in sync with the flash memory. |

---

### Address Decomposition

The 24-bit flash byte address is decoded natively to index into the direct-mapped cache architecture. With `N_SETS = 4` (2 bits) and a 4\,KiB sector size (12 bits), the address bits are mapped as follows:

| Field | Tag | Set Index | Byte Offset |
| :--- | :---: | :---: | :---: |
| **Bit Range** | `[23:14]` | `[13:12]` | `[11:0]` |
| **Width** | 10 bits | 2 bits | 12 bits |
| **Purpose** | Uniquely identifies the stored sector | Selects 1 of the 4 cache lines | Addresses specific byte inside the 4\,KiB line |

---

### Cache Controller Logic Flow

* **Cache Hit Criteria:** A cache hit is set when `Valid == 1` and the incoming `ADDRESS[23:14]` matches the stored `tag` for the indexed set.
* **Write Operations:** On a cache hit write, data is updated directly in the cache line, and the `Dirty` bit is set (`1`).
* **Eviction & Write-Back:** When a line needs to be replaced due to a cache miss, the cache asserts a write-back request to the external controller if the victim line's `Dirty` bit is set. The controller's main Finite State Machine (FSM) then sequentially handles:
  1. Writing the dirty cache line back to the physical flash memory sector.
  2. Evicting/invalidating the old line.
  3. Fetching and replacing it with the newly requested data.

--- 

### Test status

All read and write tests pass with at word-granularity transfers, where hit, miss and writeback stages are tested. Byte-addressable accesses do NOT work. Also, when the cache is disabled (`CACHE_EN=0`), all tests pass, including byte-addressable accesses.

Requires #952 to be merged first.


davideschiavone:
this PR has been moved to #980 - thanks @Patrick-Pataky

PR #975, opened 2026-06-09
Migrate configuration register generation from bash script to FuseSoC generator

StMiky:
This PR pulls in `reg-generator.py` from `x-heep-common` and integrate it in X-HEEP's peripherals `.core` files to automate the register generation process, eliminating the need to write dedicated scripts and call then in Makefile. The caching mechanism embedded in the generator ensures that the output files (RTL, software headers, docs) is only re-created when the input configuration changes.


hw/vendor/xheep/common/util/reg-generator/reg-generator.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+All paths are relative to `files_root`, which FuseSoC sets to the directory containing your `.core` file.
+
+## Parameters
+
+| Parameter          | Required | Description |
+|--------------------|----------|-------------|
+| `name`             | yes      | Name of the register block (used to derive output filenames). |
+| `config`           | yes      | Path to the HJSON register description, relative to `files_root`. A `.tpl` extension triggers Mako template rendering before passing the file to `regtool`. |
```
davidmallasen: Clarify that if you're using mcu-gen this should be the non-templated path (correct me if I'm wrong)

hw/vendor/xheep/common/util/reg-generator/reg-generator.md
```diff
@@ -0,0 +1,151 @@
...
+      - /absolute/path/to/rtl/my_peripheral_reg_top.sv
+    file_type: systemVerilogSource
+targets:
+  default:
+    filesets: [rtl]
+```
+
+## Template-based HJSON
+
+When `config` has a `.tpl` extension, the generator first renders it as a [Mako](https://www.makotemplates.org/) template before passing it to `regtool`. This is useful for parametric register maps whose field count or addresses depend on design-time constants. All `parameters` key-value pairs (e.g, `<others>`) entries are forwarded to the template as variables, plus a pre-computed `version_hex` (a `0x00MMmmPP` hex string derived from the target core's semantic version).
```
davidmallasen, line 117: Clarify use with mcu-gen

waiver_v5.core
davidmallasen, line 1: Is this dropping support for verilator 4 or a remnant of something previous?

davidmallasen (approved):

StMiky:
I clarified the external template render use case. I also fixed outdated python inclusion paths in the docs `conf.py` that were causing errors during the build. We should be good to go now.

PR #977, opened 2026-06-11
Revendor xheep_gen

LuigiGiuffrida98:
This PR integrates into X-HEEP latest edits to xheep_gen classes.

Depends on https://github.com/x-heep/xheep_gen/pull/5 and https://github.com/x-heep/xheep_dma/pull/10


hw/vendor/waiver/lint/spi_device.vlt
davidmallasen, line 1: Not needed, it was fixed in https://github.com/x-heep/x-heep/pull/1033

hw/ip/power_manager/rtl/power_manager.sv.tpl
```diff
@@ -4,11 +4,11 @@
 
 <%
     memory_ss = xheep.memory_ss()
-    external_domains = xheep.get_base_peripheral_domain().get_power_manager().get_external_domains()
+    external_domains = xheep.get_base_peripheral_domain().get_peripheral("power_manager").get_external_domains()
```
davidmallasen, line 7: I like this way of accessing all the peripherals much better. In the new peripheral_domain class I see there is still a get_power_manager and such for other peripherals, maybe we should remove those to simplify things https://github.com/x-heep/xheep_gen/blob/7327dd9f83deb1fa166f78e5453993359f7d2087/peripherals/peripheral_domain.py#L231 . If you agree because you're also not using them in X-ALP I can remove them

hw/ip/pdm2pcm/data/pdm2pcm.hjson.tpl
```diff
@@ -81,7 +81,8 @@
     }
 
 <%
-  pdm2pcm = xheep.get_user_peripheral_domain().get_pdm2pcm()
+  _dom = xheep.get_user_peripheral_domain()
```
davidmallasen, line 84: This is to allow for the `user_peripheral_domain` to not exist in the future?

hw/core-v-mini-mcu/core_v_mini_mcu.sv.tpl
```diff
@@ -7,10 +7,10 @@
 %>
 
 <%
-  dma = xheep.get_base_peripheral_domain().get_dma()
+  dma = xheep.get_base_peripheral_domain().get_peripheral("dma")
   memory_ss = xheep.memory_ss()
   user_peripheral_domain = xheep.get_user_peripheral_domain()
-  dma = xheep.get_base_peripheral_domain().get_dma()
+  dma = xheep.get_base_peripheral_domain().get_peripheral("dma")
```
davidmallasen, line 13: Duplicated (also in main, probably a mistake), can you delete one of the two?

configs/general.py
```diff
@@ -60,20 +61,20 @@ def config():
 
     system.set_debug_ss(DebugSS(has_spi_slave=1))
 
-    address_map = AddressMap()
-    address_map.add_region(
-        AddressRegion("debug", start_address=0x10000000, length=0x00100000)
+    base_peripherals = AddressRegion(
```
davidmallasen: Why save them into a variable instead of doing it as it was before?

configs/general.py
```diff
@@ -123,8 +124,8 @@ def config():
     user_peripheral_domain.add_peripheral(UART(0x00080000))
 
     # Add the peripheral domains to the system
-    system.add_peripheral_domain(base_peripheral_domain)
```
davidmallasen, line 126: Undo

davidmallasen (changes requested):
Thanks @LuigiGiuffrida98. I added a couple of minor comments, and please update the docs :)

hw/ip/power_manager/rtl/power_manager.sv.tpl
LuigiGiuffrida98: yes please, I forgot to remove them in the cleaning

hw/ip/pdm2pcm/data/pdm2pcm.hjson.tpl
LuigiGiuffrida98: It's mainly to avoid having a super long "xheep.get_user_peripheral_domain().get_peripheral("pdm2pcm")

configs/general.py
LuigiGiuffrida98: In a previous version I needed to pass this the peripheral subsystem, reverting it

PR #981, opened 2026-07-01
Vendor obi and use typedefs

LuigiGiuffrida98:


hw/core-v-mini-mcu/include/xheep_obi_pkg.sv
```diff
@@ -7,20 +7,68 @@
  *
  */
 
+`include "obi/typedef.svh"
+
+
 package xheep_obi_pkg;
 
+  /// The config type for OBI bus optional fields.
   typedef struct packed {
```
StMiky, line 16: Isn't this already defined in `obi_pkg.sv`? Can't we just include that package here?

hw/core-v-mini-mcu/include/xheep_obi_pkg.sv
```diff
@@ -7,20 +7,68 @@
...
+      AUserWidth: 0,
+      WUserWidth: 0,
+      RUserWidth: 0,
+      MidWidth: 0,
+      AChkWidth: 0,
+      RChkWidth: 0
+  };
 
+  /// The OBI bus config type.
   typedef struct packed {
```
StMiky, line 50: Same as before, isn't this already defined in `obi_pkg.sv`?

hw/core-v-mini-mcu/peripheral_subsystem.sv.tpl
```diff
@@ -651,25 +657,68 @@ module peripheral_subsystem #(
   assign ddr_i = {ddr_rcv_3_i, ddr_rcv_2_i, ddr_rcv_1_i, ddr_rcv_0_i};
   assign {ddr_snd_3_o, ddr_snd_2_o, ddr_snd_1_o, ddr_snd_0_o} = ddr_o;
 
+  obi_req_t serial_link_writer_req, serial_link_reader_req;
```
StMiky, line 660: Does this mean the we are only migrating the signal definitions but not the bus primitives? If PULP's bus primitives do not support the full OBI data types, why are we even switching to those?

hw/vendor/patches/pulp_platform_obi/0001-parametrize-optional-field-widths.patch
StMiky, line 1: Missing "e" in "parametrize" in the file name. Also why do we need this? We never use the optional field in X-HEEP

hw/vendor/waiver/lint/obi.vlt
StMiky, line 1: Is this really from OHW? If not, change the header. Also, I would narrow down the waived error messages as I guess most of them are bits being unused, not entire signals. If the entire `core_instr_resp_i` is unused, it should be flagged as an error I believe...

hw/vendor/xheep/dma/rtl/dma.sv
StMiky, line 1: Please add a link to all the external PRs this one depends on in the description, so we can check those as well

hw/vendor/xheep/dma.vendor.hjson
```diff
@@ -7,7 +7,7 @@
 
   upstream: {
     url: "https://github.com/x-heep/xheep_dma.git",
-    rev: "90416566f63c94eeec4966f10db697b6128e27eb",
+    rev: "update-obi",
```
StMiky, line 10: I know it's a work in progress, but use the commit hash instead of the branch name as the revision

hw/vendor/xheep/spi.vendor.hjson
```diff
@@ -7,7 +7,7 @@
 
   upstream: {
     url: "https://github.com/x-heep/xheep_spi.git",
-    rev: "3e741db72563bfa1b904aa90a08011f7e1e91c97",
+    rev: "update-obi",
```
StMiky, line 10: Commit hash instead of branch name

StMiky (changes requested):

PR #985, opened 2026-07-08
 Add execute from boot rom mode

alessionaclerio22:
This PR introduces a new boot mode where the CPU executes directly from an external device. A new `execute_from_ext_i` pin takes priority over `boot_sel_i`/`execute_from_flash_i`: when asserted, `boot_rom.S` jumps to the address held in a new `EXT_BOOT_ADDRESS` SOC_CTRL register (defaulting to the `ext_slaves` base address).

# Usage

1. Build the app for the external device and bake it into the example model
`make app-ext-rom PROJECT=<app>`

2. Recompile the simulation model (the ROM content is compiled into it)
`make verilator-build`

3. Run — no firmware file needed
`make verilator-run SIM_ARGS="+execute_from_ext=1"`

# Testbench boot modes

| `execute_from_ext_i` | `boot_sel_i` | `execute_from_flash_i` | boot procedure |
|---|---|---|---|
| 1 | X | X | Execute from External Device |
| 0 | 0 | X | JTAG/hex load (default) |
| 0 | 1 | 1 | SPI Flash Execution |
| 0 | 1 | 0 | SPI Flash Loading |


davideschiavone:
@alessionaclerio22 - what is the purpose of this PR? is there a specific application for it or is it of general interest? cause if the first one, we can simply add a mode where it jumps to a specific address, and then you map wherever the new bootrom with your specific things outside x-heep, and x-heep would just support a special JUMP to an address different than 0x180 -

davideschiavone (changes requested):
see previous comment

PR #990, opened 2026-07-10
Fg/dev sram

LuigiGiuffrida98:


flague:
- Support memory scrubbing in `reliable` mode and extend memories to store ECC.
- (TODO) Add intermediate `reliable` mode with non-triplicated bus but ECC on memories

flague:
- Create a reliability class with multiple supported flavors (bus and memory can be made redundant independently now)

PR #993, opened 2026-07-20
Linker script configuration support in Python

mohasnik:
Adds Python-side linker script configuration support for stack and heap sizes.

Users can now configure these sizes using Python scripts


configs/general.py
davidmallasen, line 1: Change it in all the other configuration files inside the `configs` folder, and remove it from the python unsupported hjson as well

configs/general.py
```diff
@@ -40,15 +40,21 @@
...
 
 def config():
     system = XHeep(BusType.onetoM)
     system.set_cpu(cv32e20(rv32e=False, rv32m="RV32MSlow"))
 
     memory_ss = MemorySS()
-    memory_ss.add_ram_banks([32] * 2)
-    memory_ss.add_linker_section(LinkerSection.by_size("code", 0, 0x00000E800))
-    memory_ss.add_linker_section(LinkerSection("data", 0x00000E800, None))
+    memory_ss.add_ram_banks([256] * 12)
```
davidmallasen: Don't change the configuration. It should just use the new system instead of the old one. To make sure, use the compare-mcu-gen makefile target: https://github.com/x-heep/x-heep/blob/10755cfb51a89e24a4ef1c68b277426c52549d14/Makefile#L381

configs/general.py
```diff
@@ -40,15 +40,21 @@
...
+    memory_ss.add_linker_section(LinkerSection.by_size("code", 0, 0x000040000))
+    memory_ss.add_linker_section(LinkerSection("data", 0x000040000, None))
+
+    system.set_linker_script_config(
+        LinkerScriptConfig(stack_size=0x4000, heap_size=None)
```
davidmallasen: Use the same configuration as was set before by the hjson. The functionality should be unchanged

util/xheep_gen.vendor.hjson
```diff
@@ -6,8 +6,8 @@
   target_dir: "xheep_gen",
 
   upstream: {
-    url: "https://github.com/x-heep/xheep_gen.git",
-    rev: "95efdefe2b8d38d9752fa9ba59666dbe5faf212e",
+    url: "https://github.com/mohasnik/xheep_gen.git",
```
davidmallasen: To be updated to the upsream repo once it's merged

davidmallasen (changes requested):
Thanks @mohasnik ! Also check the failing CI

davidmallasen:
Modify the docs with the new updated changes @mohasnik

davidmallasen (approved):

PR #1000, opened 2026-07-24
Feat/librelane flow

nathanchandanson:
The goal of this PR is to add two ways of making a chip of X-Heep: one macro and one chip flow.
Everything is explained in the `ImplementASIC.md` file.

The folder `hw/asic/ihp-sg13g2` contains all the files to run both flows. When running `make mcu-gen`, all the config files (Librelane, pdn, sdc, macro placement,...) as well as wrappers will be generated according to the config chosen.

I also had to modify `hw/system`, especially the `pad_ring` and the `x_heep_system` to use and instantiate the power pins/pads.

Support for power pins/pads was added to `xheep_gen` with the creation of the `PinPower` and other classes.

Note that this PR relies on [this PR](https://github.com/x-heep/edalize/pull/7) to `edalize` to implement the librelane tool.


davideschiavone:
hello @nathanchandanson - very nice PRs - would you be available for a quick call? I guess the review could be done together on a few points while talking :) if you can, reach out on the X-HEEP matrix channel at https://matrix.to/#/#x-heep:matrix.org

configs/pad_cfg.py
```diff
@@ -77,6 +77,19 @@ def config(xheep: XHeep) -> PadRing:
         Output("ddr_snd_3"),
     ]
 
+    ##############################################
```
davideschiavone: move this to pad_cfg_openlane.py

hw/asic/ihp-sg13g2/rtl/if_xif_structs_pkg.sv.tpl
```diff
@@ -0,0 +1,101 @@
+<%
```
davideschiavone: this is redundant - @LuigiGiuffrida98

configs/pad_cfg.py
davideschiavone: @nathanchandanson - can you add a config/openlane.ihp130.py/hjson

hw/system/pad_ring.sv.tpl
```diff
@@ -47,6 +48,30 @@ module pad_ring (
             % endfor
         `endif
     %endif
+`ifdef ASIC_SYNTHESIS
+    % if len(power_pads) > 0:
+        `ifdef USE_POWER_PINS
+        <%
```
davideschiavone: make a copy to hw/asic/ihp etc... @nathanchandanson

hw/system/x_heep_system.sv.tpl
```diff
@@ -95,7 +96,35 @@ module x_heep_system #(
     // External SPC interface
     output logic [core_v_mini_mcu_pkg::DMA_CH_NUM-1:0] dma_done_o,
 
-    % for pad in xheep.get_padring().pad_list:
+`ifdef ASIC_SYNTHESIS
```
davideschiavone: let's discuss with @JuanSapriza

util/xheep_gen/pads/pin.py
```diff
@@ -112,3 +112,54 @@ def __init__(self, name, module=None, attributes=None):
         super().__init__(name, module, attributes)
         self.iocell = iocell_a.copy()
         self.bondpad = bondpad_a.copy()
+
```
davideschiavone: @JuanSapriza

davideschiavone (changes requested):

nathanchandanson:
Hi, I modified the things we talked about.
Now my ASIC wrappers are `x_heep_system` slightly modified. Thus, I could revert the changes I made in the `hw/x_heep_system` folder.
I also changed a bit the folders/files so that it is a bit clearer to understand.
And I duplicated the config files with a different name so that my setup is more reproducible.

Let me know if you want me to modify some other things.

util/xheep_gen/pads/pin.py
JuanSapriza: Hello @nathanchandanson thanks for this PR! Cool work! 
I believe the pad definitions should be inside `hw/asic/ihp-sg12g2/pad_definitions.py` (or similar), because it's pdk specific. It doesnt hurt to have them here, but i think logically it would make more sense to move it.  

Same applies for the Cell definitions. 

A (fake) example of a definitions file (in case its useful) in which you can see how to update the default cells and pin types. 

```python
from x_heep_gen.pads import cell
from x_heep_gen.pads.pin import *
from x_heep_gen.pads.pad import *

# Create the PDK pads
iocell_aioVss   = Cell( width=xx, height=xx, name="CELL NAME XXX", rtl_wrapper="u_pad_cell_analog_io_vss",     connections=["vss"] )
iocell_aHiZ     = Cell( width=xx, height=xx, name="CELL NAME YYY",   rtl_wrapper="u_pad_cell_analog_hiz",        connections=["io"] )

# PDK physical cells
aPrcut          = Cell( width=xx, height=xx, name="PR_CUT_ANALOG_NAME"  )
dPrcut          = Cell( width=xx, height=xx, name="PR_CUT_DIGITAL_NAME"  )

# Update the standard cells to match the dimensions and cell name from the PDK
cell.bondpad_a.update(  dimension=Dimension( width=xx, height=xx), name="CELL NAME ZZZ" )
cell.iocell_d.update(   dimension=Dimension(width=xx, height=xx), name="CELL NAME WWW" )

# Just a dummy cell to use as a filler when you want to leave a space
pad_skip        = Cell(name="", width=0, height=0)

# Define custom Pin classes. These classes are then assigned to pins and they will inherit the needed io cells and bondpads

class Space(Physical):
    """
    To leave empty space between pads
    """
     def __init__(self, width):
        super().__init__("space", iocell=pad_skip.copy(), bondpad=None)
        self.iocell.dimension.width = width

class PinSupply(Pin):
    """
    Represents a generic supply pin.
    """

    def __init__(self, name, attributes=None):
        super().__init__(name, attributes=attributes)

class DVdd(PinSupply):
    """
    Represents a digital Vdd supply pin.
    """

    def __init__(self, name, attributes=None):
        super().__init__(name, attributes)
        self.iocell = iocell_dVdd.copy()
        self.bondpad = bondpad_d.copy()

class DVddIO(PinSupply):
    """
    Represents a digital I/O Vdd supply pin.
    """

    def __init__(self, name, attributes=None):
        super().__init__(name, attributes)
        self.iocell = iocell_ioVdd.copy()
        self.bondpad = bondpad_d.copy()
. . . etc etc etc

```

nathanchandanson: Hi, I  started adding an `ihp_pad_configuration.py` inside `hw/asic/ihp-sg13g2`.
The issue I'm having is to resolve where should python look for modules, because as for now `hw/asic/ihp-sg13g2` is not in the `PYTHONPATH`. Thus for now I am doing `PYTHONPATH=hw/asic/ihp-sg13g2  make mcu-gen ...` so that my `ihp_pad_configuration.py` file is included as a module, as  it is needed for generating all my templated files.

Do you think it is a good way to do it ? If no, how would you do it ?

davideschiavone: @JuanSapriza , @davidmallasen

JuanSapriza: Good point, I had the same issues. 

For the pad_config file (where you define which pad goes where) I was using:
```python
import sys, os
sys.path.append(os.path.dirname(__file__))
from cheep_pad_defs import *
```

This way you dont need to add it manually, it's added automatically when you choose the configuration file to be used. 

Then, on template files I could simply use:
```python
<%!
    from x_heep_gen.pads.pin import Input, Output, Inout, PinDigital, Asignal
    from cheep_pad_defs import ASignal_HiZ, AVss, AVddIO, AVssIO
%>
```

This was generally enough, but if there is any case in which this doesnt work we can think of an alternative solution. 

@davidmallasen maybe on DHB you are managing this in a different way?

nathanchandanson: If I do so, then my `pad_config` file needs to be inside `hw/asic/ihp-sg13g2` and not in `configs` right ?
Or if I let the config file in `configs` I can do:
```
sys.path.append(
    os.path.join(os.path.dirname(__file__), "..", "hw", "asic", "ihp-sg13g2")
)
```

@davideschiavone @JuanSapriza Which one seems better to you ?

JuanSapriza: i dont have a very strong opinion, but i tend to believe that the most tidy option would be to keep what is PDK-only-dependant in the pdk folder, and what is chip-dependant on the configs folder (i.e. if we have two chips on the same PDK the pad defs would not change, but the configs would). 

TLDR, I would leave the configs in configs and do `os.path.join(os.path.dirname(__file__), "..", "hw", "asic", "ihp-sg13g2")`

davidmallasen: In DHB we didn't have this issue so far so I haven't tried any fix for this. The one you propose @JuanSapriza sounds good. For the `pad_config` file you can put it wherever you want, as you can set the path with the `PADS_CFG` variable to the makefile

nathanchandanson: Great thank you I will keep it inside `configs` then.

nathanchandanson: Hello, I continued a bit on the PR.
When doing `make mcu-gen`  it tries to generate my LibreLane templates, but it fails because they specifically need the `pad_cfg_librelane.py` config file and the default one does not contain `sys.path.append(os.path.join(os.path.dirname(__file__), "..", "hw", "asic", "ihp-sg13g2"))`.

Thus I would like to exclude the ASIC files from being generated when the goal is not to implement an ASIC.

@JuanSapriza @davidmallasen Is there a way to exclude specific template files when running `make mcu-gen` ?
I though about doing something like this at the beginning of my ASICs templated files:
```
<%
    if not context.get(GENERATE_ASIC_FILES):
        return STOP_RENDERING
%>
```
However, I cannot find where I should put my `GENERATE_ASIC_FILES` variable definition for it to be passed to the templates. 

Is it the good way to do it ? If yes where would you put the `GENERATE_ASIC_FILES`, and if no how would you do it ?

davidmallasen: We could use the `STOP_RENDERING` from mako that you suggest with the `context`. For that, I would use the `TARGET` variable in the makefile that now is used for `sim` or different FPGA targets

nathanchandanson: I did this in the `mcu_gen.py`:
```
diff --git a/util/xheep_gen/mcu_gen.py b/util/xheep_gen/mcu_gen.py
index 925e8c0..f2e1476 100755
--- a/util/xheep_gen/mcu_gen.py
+++ b/util/xheep_gen/mcu_gen.py
@@ -17,6 +17,7 @@ from mako.template import Template
 import load_config
 from xheep import BusType
 from cpu.cpu import CPU
+import os


 # ANSI color codes for pretty printing
@@ -145,6 +146,9 @@ def generate_xheep(args):
             + " bytes."
         )

+    # Extract the target environment variable
+    impl_target = os.getenv("TARGET")
+
     kwargs = {
         "xheep": xheep,
         "debug_start_address": debug_start_address,
@@ -159,6 +163,7 @@ def generate_xheep(args):
         "plic_used_n_interrupts": plic_used_n_interrupts,
         "plit_n_interrupts": plit_n_interrupts,
         "interrupts": interrupts,
+        "impl_target": impl_target,
     }

     return kwargs
```
And then inside the templates it is possible to use the `impl_target` variable:
```
<%
    if impl_target != "asic_ihp":
        return STOP_RENDERING
%>
```
Now I can selectively generate or not the templates when doing `make mcu-gen TARGET=asic_ihp`.

@davidmallasen Is it what you had in mind ?

davidmallasen: Yep, something like that

nathanchandanson:
Hello,

I just finished reworking a bit everything we discussed.
Now everything that is PDK specific is inside `hw/asic/ihp-sg13g2` and config files are inside the `configs` folder.

The only notable change is that I modified a bit the `mcu_gen.py` script so that we can get the `TARGET` environment variable from inside the templates.
I used it to avoid generating my ASIC files whenever they are not needed, as they require a specific configuration.

@davideschiavone Let me know if you see other changes I have to make.

davideschiavone:
> Hello,
> 
> I just finished reworking a bit everything we discussed. Now everything that is PDK specific is inside `hw/asic/ihp-sg13g2` and config files are inside the `configs` folder.
> 
> The only notable change is that I modified a bit the `mcu_gen.py` script so that we can get the `TARGET` environment variable from inside the templates. I used it to avoid generating my ASIC files whenever they are not needed, as they require a specific configuration.
> 
> @davideschiavone Let me know if you see other changes I have to make.

Thanks @nathanchandanson - I will let @LuigiGiuffrida98 have the final merge as he's going to take over the project with open source tool! but looks good to me - thanks a lot for your contribution!!!

nathanchandanson:
Great perfect !

I remain available in case you need info about what I did.
Also don't hesitate to identify me in case there are issues or questions.

PR #1032, opened 2026-09-17
add HLS Vitis example

davideschiavone:
## Description

Adding an example of HLS accelerator integration extention.
The module is simple as it doesnt contain any memory - thus it can be simulated by vendor neutral tools.
It requires Vitis to generate the Verilog, and its verilog can be simulated with Verilator and other EDA simulation tools.
It can be instantiated in the Pynq-z2 top level to extend X-HEEP.

## Checklist

- [x] HLS accelerator synthesizes and generates Verilog
- [x] Verilog is simulated within X-HEEP with EDA sim tools
- [ ] FPGA bitstream builds correctly
- [ ] FPGA bitstream runs correctly
- [ ] Documentation updated


core-v-mini-mcu.core
```diff
@@ -392,6 +409,18 @@ targets:
...
           - '--x-assign unique'
           - '--x-initial unique'
+          # Vitis HLS-generated RTL (e.g. hw/fpga/hls/vitis/dot_product) uses
+          # zero-time delay statements in a couple of small pipeline flow
+          # control modules, purely as a simulation-only race-avoidance
+          # trick with no real timing meaning. Verilator 5.x requires an
+          # explicit choice of how to handle any delay/event control it
+          # sees; --no-timing tells it to treat them as zero-time, which is
+          # exactly what they already are.
+          - '--no-timing'
```
davideschiavone: tbd with @StMiky

